PR #6, opened 2018-12-18
[near-contract-helper] api to create an account (still hardcoding keys)

janedegtiareva:


app.js
```diff
@@ -121,14 +134,45 @@ router.post('/contract/view/:name/:methodName', async ctx => {
...
 });
 
+/**
+ * Create a new account. Generate a throw away account id (UUID).
+ */
+router.post('/account', async ctx => {
+    // TODO: this is using alice account to create all accounts. We may want to change that.
+    const nonce = await getNonce(ctx, defaultSender);
+    ensureTrue(nonce);
+    const newAccountName = uuidV4();
```
vgrichina: This is totally not blocking this PR, but I think it'll be nice to have readable word combinations as account names instead, using e.g. using this package:
https://www.npmjs.com/package/rword

app.js
```diff
@@ -121,14 +134,45 @@ router.post('/contract/view/:name/:methodName', async ctx => {
...
+    console.log("Creating new account " + newAccountName);
+
+    // TODO: unhardcode key
+    const createAccountParams = {
+        'nonce': nonce,
```
vgrichina: no need to use quotes on property names

app.js
```diff
@@ -121,14 +134,45 @@ router.post('/contract/view/:name/:methodName', async ctx => {
...
+
+    const resp = await submit_transaction_rpc(client, "create_account", createAccountParams);
+    checkError(ctx, resp);
+    ctx.body = resp.result;
+});
+
+/*
+ * Helper to encode the account name in the format which is used in rpcs.
+ */
+async function encodeAccountNameForRpc(plainTextName){
```
vgrichina: this function doesn't do anything. If you think `hash` isn't clear enough name – just rename it.

app.js
```diff
@@ -121,14 +134,45 @@ router.post('/contract/view/:name/:methodName', async ctx => {
...
+        'public_key': hardcodedKey.public_key,
+    };
+
+    const resp = await submit_transaction_rpc(client, "create_account", createAccountParams);
+    checkError(ctx, resp);
+    ctx.body = resp.result;
+});
+
+/*
+ * Helper to encode the account name in the format which is used in rpcs.
```
vgrichina: This comment doesn't provide more information than method name. Let's avoid comments which don't provide additional info.

app.js
```diff
@@ -44,6 +47,12 @@ const checkError = (ctx, response) => {
     }
 };
 
+const ensureTrue = (boolValue) => {
+    if (!boolValue) {
+        ctx.throw(400, `Internal error`);
```
vgrichina: `ctx` isn't available here.

app.js
```diff
@@ -121,14 +134,45 @@ router.post('/contract/view/:name/:methodName', async ctx => {
...
     ctx.body = response.result;
 });
 
+/**
+ * Create a new account. Generate a throw away account id (UUID).
+ */
+router.post('/account', async ctx => {
+    // TODO: this is using alice account to create all accounts. We may want to change that.
+    const nonce = await getNonce(ctx, defaultSender);
+    ensureTrue(nonce);
```
vgrichina: 1. You can use `ctx.assert(nonce)` if needed, no need for custom helper.
2. Why is it needed? Are there any cases when `view_account` doesn't return `nonce` without error?

app.js
vgrichina: Internal error is `500`. `400` has specific meaning:
https://httpstatuses.com/400

app.js
janedegtiareva: I wanted to have a wrapper so that, if we decide to stop hashing names as per conversation yesterday, we don't have to change in 10 places. hash is a good name for readability, so I want to keep that as a separate thing. this said, I don't care enough to insist on this.

app.js
janedegtiareva: This is more for code readability - it spells out that we are specifically looking for non empty nonce.

app.js
janedegtiareva: removed function

app.js
vgrichina: we don't use `hash` for anything else. We can rename it to something like `accountHash` to be more specific, I'm ok with that.

PR #8, opened 2018-12-19
[near-contract-helper] construct useful client repsonse from /account

janedegtiareva:


app.js
```diff
@@ -138,26 +138,36 @@ router.get('/account/:name', async ctx => {
...
         amount: newAccountAmount,
-        public_key: hardcodedKey.public_key,
+        public_key: accountKey.public_key,
     };
 
-    const resp = await submit_transaction_rpc(client, "create_account", createAccountParams);
-    checkError(ctx, resp);
-    ctx.body = resp.result;
+    const transactionResponse = await submit_transaction_rpc(client, "create_account", createAccountParams);
+    checkError(ctx, transactionResponse);
```
vgrichina: This method doesn't take `ctx` anymore:
https://github.com/nearprotocol/near-contract-helper/pull/7/files

vgrichina: `submit_transaction_rpc` also changed.

app.js
```diff
@@ -138,26 +138,36 @@ router.get('/account/:name', async ctx => {
...
+
+    // transactionResponse does not contain useful information, so construct a custom response
+    // TODO: is it fine to return private key here? 
+    const clientResponse = {
+        account_name: newAccountName,
```
vgrichina: let's use camelCase in our responses.

app.js
```diff
@@ -138,26 +138,36 @@ router.get('/account/:name', async ctx => {
...
     };
 
-    const resp = await submit_transaction_rpc(client, "create_account", createAccountParams);
-    checkError(ctx, resp);
-    ctx.body = resp.result;
+    const transactionResponse = await submit_transaction_rpc(client, "create_account", createAccountParams);
+    checkError(ctx, transactionResponse);
+
+    // transactionResponse does not contain useful information, so construct a custom response
+    // TODO: is it fine to return private key here?
```
vgrichina: we don't need to return it, we need to record it as signing is handled on our side.

app.js
```diff
@@ -138,26 +138,36 @@ router.get('/account/:name', async ctx => {
 
 /**
  * Create a new account. Generate a throw away account id (UUID).
```
vgrichina: Let's better do API docs in a separate doc using https://apiblueprint.org
I think ultimately it's far more clear if you can see possible example responses, etc

janedegtiareva: Agreed. It will be easier to do more robust docs from these breadcrumbs though, so let's keep them for now.

app.js
```diff
@@ -177,21 +177,32 @@ router.get('/account/:name', async ctx => {
...
     };
 
-    ctx.body = await submitTransaction("create_account", createAccountParams);
+    const transactionResponse = await submitTransaction("create_account", createAccountParams);
+    checkError(transactionResponse);
+
+    // transactionResponse does not contain useful information, so construct a custom response
+    // TODO: record key info
+    const clientResponse = {
+        accountName: newAccountName,
```
vgrichina: let's also return `accountHash`, cause now we need to use hashes when passing params to contract method calls.

PR #13, opened 2019-01-09
Use JSON to serialize arguments

vgrichina:


janedegtiareva:
The change LGTM! Are you going to do the change in nearlib or should I?

vgrichina:
> The change LGTM! Are you going to do the change in nearlib or should I?

I'm ok with either, let me know if you want to do the change.

PR #14, opened 2019-01-09
[accounts] rename account id field

janedegtiareva:


app.js
```diff
@@ -184,7 +184,7 @@ router.post('/account', async ctx => {
 
     const createAccountResponse =
         await account.createAccountWithRandomKey(newAccountName, 1, defaultSender);
-    createAccountResponse['accountName'] = newAccountName;
+    createAccountResponse["account_id"] = newAccountName;
```
vgrichina: nit: `createAccountResponse.account_id` or even better:

```
ctx.body = {...createAccountResponse, account_id: newAccountName};
```

PR #24, opened 2019-02-12
Modify contract helper to conform to the new way of deploying contract

bowenwang1996:
We should also find way to securely store the keypair for a new account.

### DIFF
--- a/app.js
+++ b/app.js
@@ -76,13 +76,12 @@ router.post('/account', async ctx => {
     // TODO: this is using alice account to create all accounts. We may want to change that.
     const body = ctx.request.body;
     const newAccountId = body.newAccountId;
-    // TODO: we should find a way to store the key pair for the new account
+    const newAccountPublicKey = body.newAccountPublicKey;
     await near.waitForTransactionResult(
-        await account.createAccount(newAccountId, defaultKey.getPublicKey(), NEW_ACCOUNT_AMOUNT, defaultSender));
+        await account.createAccount(newAccountId, newAccountPublicKey, NEW_ACCOUNT_AMOUNT, defaultSender));
     const response = {
         account_id: newAccountId
     };
-    keyStore.setKey(newAccountId, defaultKey);
     ctx.body = response;
 });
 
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -77,12 +76,13 @@ router.post('/account', async ctx => {
     // TODO: this is using alice account to create all accounts. We may want to change that.
     const body = ctx.request.body;
     const newAccountId = body.newAccountId;
-    const newAccountPublicKey = body.newAccountPublicKey;
+    // TODO: we should find a way to store the key pair for the new account
```
vgrichina: I don't get this. This method previously allowed us to create and store key pair client side and just send a public key that should have access to new account. Why do we have to use default key now?

package.json
```diff
@@ -25,7 +25,7 @@
     "koa-json-body": "^5.3.0",
     "koa-logger": "^3.2.0",
     "koa-router": "^7.4.0",
-    "nearlib": "0.2.1",
+    "nearlib": "../nearcore/nearlib",
```
vgrichina: this shouldn't be committed

app.js
bowenwang1996: Because I don't see a way to store the keypair in the keystore. The keypair will be used to sign transaction from that newly created account, which is what `deployContract` depends on.

bowenwang1996: Unless we want to pass the private key here, but I feel like there is some security issue with that

vgrichina: there is no need to store it in server keystore. This is used by NEARStudio, it stores key client-side.

bowenwang1996: But the signer needs to look up the private key to sign transactions

bowenwang1996: I think I now understand how this things works. Reverted my change

janedegtiareva: We will need to have nearstudio use nearlib directly to deploy instead of calling contract helper, since contract helper doesn't have this contract-specific key.

package.json
bowenwang1996: Fixed

PR #25, opened 2019-02-28
automation fixups

azban:


app.js
```diff
@@ -91,7 +91,7 @@ app
     .use(router.allowedMethods());
 
 if (!module.parent) {
-    app.listen(process.env.PORT || 3000);
+    app.listen(process.env.NEAR_CONTRACT_HELPER_PORT || 3000);
```
vgrichina: hmm, is it hard to set different env var values for different processes?

azban: it's not a huge deal, but prefixing helps avoid conflicts with system variables and being explicit helps when someone is exporting many env variables, and can also help avoid conflicts. also, have sorta picked up the convention from configuration tools that will allow overrides from environment variables. if you think this is annoying, i can change it back.

vgrichina: I think we can do this:

`app.listen(process.env.NEAR_CONTRACT_HELPER_PORT || process.env.PORT || 3000);`

this will be non-annoying when used locally and solve problems you mention for automated envs

PR #28, opened 2019-04-30
Account recovery

vgrichina:


app.js
```diff
@@ -93,6 +59,48 @@ router.post('/account', async ctx => {
     ctx.body = response;
 });
 
+const password = require('secure-random-password');
+const models = require('./models');
+const FROM_PHONE = '+14086179592';
```
janedegtiareva: what is this phone number? comment please

app.js
```diff
@@ -93,6 +59,48 @@ router.post('/account', async ctx => {
...
+router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
+    const accountId = ctx.params.accountId;
+    const phoneNumber = ctx.params.phoneNumber;
+
+    const securityCode = password.randomPassword({ length: 6, characters: password.digits });
```
janedegtiareva: define 6 as a const? (also is 6 too few chars?)

app.js
```diff
@@ -93,6 +59,48 @@ router.post('/account', async ctx => {
...
+    const [account] = await models.Account.findOrCreate({ where: { accountId, phoneNumber } });
+    await account.update({ securityCode });
+
+    const accountSid = process.env.TWILIO_ACCOUNT_SID;
```
janedegtiareva: do we need changes in readme how to set these?

app.js
vgrichina: it is a a phone number from which codes are sent, thought this is more more or less obvious. Added env variable as well, going to document that in README

app.js
vgrichina: Done

app.js
vgrichina: README done

app.js
vgrichina: done

PR #30, opened 2019-06-05
Account recovery

vgrichina:


app.js
```diff
@@ -95,16 +95,23 @@ router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
...
+const bs58 = require('bs58');
 const verifySignature = (nearAccount, securityCode, signature) => {
     const hasher = crypto.createHash('sha256');
     hasher.update(securityCode);
+    const hash = hasher.digest();
     const publicKeys = nearAccount.public_keys.map(key => Buffer.from(key));
-    return publicKeys.some(publicKey => nacl.sign.detached.verify(hasher.digest(), Buffer.from(signature), publicKey));
+    const helperPublicKey = bs58.decode(defaultKey.publicKey);
+    if (!publicKeys.some(publicKey => publicKey.equals(helperPublicKey))) {
+        throw Error(`Account ${nearAccount.account_id} doesn't have helper key`);
```
janedegtiareva: change helper to contract helper to be more explicit?

vgrichina: it's a good question, I feel like we have to rename whole thing, cause it's not about contracts anymore. I wonder if it should be `near-account-helper`?

PR #31, opened 2019-06-12
Docker and env variables spec

nearmax:


app.js
```diff
@@ -40,7 +40,7 @@ if (process.env.NEAR_CONTRACT_HELPER_PUBLIC_KEY && process.env.NEAR_CONTRACT_HEL
 const defaultKey = new KeyPair(publicKey, secretKey);
 const keyStore = new InMemoryKeyStore();
 keyStore.setKey(defaultSender, defaultKey);
-const localNodeConnection = new LocalNodeConnection(process.env.INTERNAL_NODE_HOST || 'http://localhost:3030');
+const localNodeConnection = new LocalNodeConnection(process.env.NODE_URL || 'https://studio.nearprotocol.com/devnet');
```
vgrichina: why change default value?

app.js
```diff
@@ -141,7 +141,7 @@ app
     .use(router.allowedMethods());
 
 if (!module.parent) {
-    app.listen(process.env.INTERNAL_CONTRACT_HELPER_PORT || process.env.PORT || 3000);
+    app.listen(3000);
```
vgrichina: `PORT` is still very useful for running locally

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN echo 'deb https://dl.yarnpkg.com/debian/ stable main' | tee /etc/apt/sources.list.d/yarn.list
+RUN apt-get update -qq && apt-get install -y \
+    jq \
+    nodejs
+
+
+# contract-helper
+COPY . /near-contract-helper/
+WORKDIR /near-contract-helper
+RUN npm install
```
vgrichina: this doesn't have to block, but let's start switching to `yarn` everywhere. Looks like it's faster and more reliable as well.

app.js
nearmax: The convention across all repos is that the default values are for the main setup that runs in the cloud is under our main DNS. GCloud deployment or local deployment need to override them.

app.js
nearmax: Same question as in here: https://github.com/nearprotocol/near-fiddle-api/pull/8#discussion_r293169650

Dockerfile
nearmax: Okay, let's do it after we have the first version.

vgrichina: Up to you, it might speed up your testing process )

PR #69, opened 2020-01-07
Add API call to send recovery message

vgrichina:


vgrichina:
@kcole16 I'm using placeholder copy for now, may you help with message templates?

app.js
```diff
@@ -133,6 +147,81 @@ router.post('/account/:phoneNumber/:accountId/validateCode', async ctx => {
...
+            to: email,
+            subject: `Important: Near Wallet Recovery Email for ${accountId}`,
+            text:
+`Hello ${accountId}!
+
+Use this link to recover account:
+${recoverUrl}
+
+Alternatively use this backup phrase:
+${seedPhrase}
```
Patrick1904: If we're exposing a backup phrase in the recovery email, then we may as well offer 'send backup phrase to email' option in the wallet? Currently we force user to only write down phrase, but that extra 'security' doesn't help much if we include it in email recovery

vgrichina: Hmm, I thought we don't send recovery e-mail when user has picked the option to use recovery phrase?

Patrick1904: Currently if user finishes signup flow with seed phrase or SMS, we don't send an email. But part of 'recovery management' (coming up) is user being able to enable additional recovery methods post signup. If the user enables email recovery post signup, then we'd send an email w/ magic link and seed phrase.

If we do not allow users to copy the seed phrase, or send it to their email, then we should probably not include the seed phrase in email recovery since it's sends mixed messages.

kcole16:
@vgrichina Thoughts on this?

```
Hi <account_name>,

This email contains your NEAR account recovery link. 

Keep this email safe, and DO NOT SHARE IT! We cannot resend this email.

Click below to recover your account.

<Recover Account Button> (fallback is link)
```

Omitted the seed phrase in this version for simplicity.

app.js
vgrichina: We should make it clear which security risks are involved in email-based recovery. I think generally we should steer users towards removing this message-based recovery option in "recovery management" and switch to writing down the seed phrase.

Note that we don't have same problem as ETH where your account is forever stuck with the same seed phrase, every time recovery method is set up we can use different seed phrase.

PR #86, opened 2020-02-17
refactor: improve getting-started experience

chadoh:
This mostly boils down to README improvements, but does introduce one **breaking change**:

* Rather than expecting `CREATOR_KEY` and `RECOVERY_KEY` to have a `private_key` attribute, they are now expected to have a `secret_key`.  This makes it simpler to get started, since `~/.near/validator_key.json` can be copied directly, with no key renames.

Most other noise here comes from the fact that all `.env` settings are now required – this simplifies the code. Many of these were already required but poorly documented; this documents them.

Also interesting: note that the `start` command has changed to use `sed` instead of just `cat`. This is necessary to strip end-of-line comments.  These end-of-line comments provide a much nicer onboarding experience and code-organization/maintenance scheme than placing this information in the README, and are worth the small increase in complexity of the `start` command.

Finally, the expectation established in `.env.sample` is that you will run this codebase connected to a local NEAR network, rather than expecting people to connect to production.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,55 +1,61 @@
 # near-contract-helper
-Micro-service used to create accounts.
 
-## Environment Variables
-This micro-service depends on the following environment variables:
-* `NODE_URL` -- default `https://studio.nearprotocol.com/devnet`;
-* `PORT` -- default `3000`;
-
-The above variable is used internally by the contract helper, it does not have to correspond to the external IP or DNS
-name and can link to the host machine running the Docker container.
-
-* `NODE_ENV` -- default `production`. Node.js environment (should be `production` for production use, `development` for local development)
-* `NEW_ACCOUNT_AMOUNT` -- default `10000000000`, integer;
-* `ACCOUNT_CREATOR_KEY` -- JSON key of account used to create accounts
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,55 +1,61 @@
...
+
+### Create database
+
+Follow the instructions above for creating the development database and `helper` user. Then create an `accounts_test` database using `psql`:
+
+    create database accounts_test with owner=helper;
+
+### TODO
+
+How to get the tests running?
```
chadoh: This is a question for you, @vgrichina

vgrichina: got tests running on CI here https://github.com/nearprotocol/near-contract-helper/pull/89
same setup should work locally

app.js
```diff
@@ -36,18 +36,18 @@ const recoveryKeyJson = JSON.parse(process.env.ACCOUNT_RECOVERY_KEY);
 const keyStore = {
     async getKey(networkId, accountId) {
         if (accountId == creatorKeyJson.account_id) {
-            return KeyPair.fromString(creatorKeyJson.private_key);
```
vgrichina: @chadoh we'll need to update secrets on Render based on this change before merging this PR. Alternatively we can support both `secret_key` and `private_key` to allow merging and deploying as is.

PR #88, opened 2020-02-18
feat: improve recovery method management

chadoh:
[In `near-wallet`][frontend], we want to allow managing existing recovery methods and support adding new ones. This requires a few new features here in `near-contract-helper`.

1. For each recovery method (phone, email, phrase), we want to display when that method was added to the account. At some point, we may want to store this information on-chain, but for now I've added it to the database here. The best way to manage this is a new database table, `RecoveryMethods`. To properly set up this table required a complex data migration, which I've split across four files to ensure they run sequentially.
2. Our `sendRecoveryMessage` endpoint previously assumed that it would only be called once for a given account. Now that the frontend will allow setting both a phone and an email as recovery methods, we need to update the logic so it can be called multiple times, with relevant information updated in the db each time.
3. This logic changes our data expectations – this app now expects that the `accountId` column will be unique. See the first two new migrations for a new database constraint to match this expectation.
4. We need new endpoints:
   1. for returning existing recovery methods
   2. for deleting recovery methods
   3. for updating the `phraseAddedAt` attribute – we don't call `sendRecoveryMessage` for this one
5. Make sure only an account's owner can call these endpoints

I've also added fairly comprehensive tests for all of these new endpoints, which serve both as traditional quality assurance and as documentation for how to use them. If you're reviewing, I recommend starting with the tests, to double-check my logic, especially around security.

I have added [some frontend logic](https://github.com/nearprotocol/near-wallet/pull/399) for interacting with these new endpoints. My local testing has worked well so far.

Finally, this new logic required some changes to the `verifySignature` function. This function had previously only been used by the `validateCode` endpoint, but [this endpoint is no longer used](https://github.com/nearprotocol/near-wallet/pull/437). I've removed this endpoint and the now-obsolete `ACCOUNT_RECOVERY_KEY`.

  [frontend]: https://github.com/nearprotocol/near-wallet/pull/381


app.js
```diff
@@ -56,6 +60,29 @@ app.use(async (ctx, next) => {
     await next();
 });
 
+async function verifyAccountOwnership({ ctx, accountId, securityCode, signature }) {
+    // ensure securityCode is a recent block
+    const givenBlock = Number(securityCode);
+    if (Number.isNaN(givenBlock)) ctx.throw(403);
+
+    const currentBlock = (await ctx.near.connection.provider.status()).sync_info.latest_block_height;
```
chadoh: Is there a better way to get current block height?

test/app.test.js
```diff
@@ -67,6 +83,37 @@ describe('/account/sendRecoveryMessage', () => {
...
+            .send({
+                accountId: ctx.accountId,
+                phoneNumber,
+                seedPhrase: SEED_PHRASE
+            });
+        account = await models.Account.findOne({
+            where: { accountId: ctx.accountId }
+        });
+        await account.reload();
+        expect(account.phoneNumber).toBe(phoneNumber);
```
chadoh: This test fails intermittently on this line and I'm not sure why – my best guess is a caching issue with `sequelize`. Note that I've already added `reload()` calls _in addition_ to the `findOne` lookups – I would think that both are not necessary, but in my testing I found that data would be consistently stale without both `reload()` calls. 😒 

I'm not sure how to make this test more robust than it already is, or how to restructure it to adequately test this functionality without running into caching issues. Ideas?

app.js
```diff
@@ -18,7 +18,11 @@ app.use(async function(ctx, next) {
             ctx.throw(e.response.status, e.response.text);
         }
 
-        if (e instanceof httpErrors.Forbidden) {
+        if (
+            e instanceof httpErrors.Forbidden ||
```
vgrichina: let's not use `instanceof` to check for errors, it's incredibly fragile (see e.g. https://github.com/isaacs/rimraf/commit/aa50e0298f66b3375045e8d9d86992d696a56b09)

app.js
```diff
@@ -56,6 +60,29 @@ app.use(async (ctx, next) => {
     await next();
 });
 
+async function verifyAccountOwnership({ ctx, accountId, securityCode, signature }) {
```
vgrichina: few nits:

1) it makes sense to structure this as Koa middleware, see e.g. https://github.com/nearprotocol/near-fiddle-api/blob/master/app.js#L88
2) to make it easier to reason about, it's best to make allowed cases explicit instead of making not allowed cases explicit.
3) Why use `securityCode` for smth that's not a security code anymore?

So whole method can be something like:

```
const currentBlock = (await ctx.near.connection.provider.status()).sync_info.latest_block_height;
if (givenBlock > currentBlock - 100 && givenBlock <= currentBlock) {
    const nearAccount = await ctx.near.account(accountId);
    if (!await verifySignature(nearAccount, securityCode, signature)) {
        ctx.throw(403);
    }
}
```

no need to handle all other errors as 403 as well, they already won't result in successful status

app.js
vgrichina: Don't think so

app.js
```diff
@@ -219,17 +302,20 @@ router.post('/account/sendRecoveryMessage', async ctx => {
...
     if (phoneNumber) {
-        where.phoneNumber = phoneNumber;
-    } else if (email) {
-        where.email = email;
+        account.update({ phoneNumber, phoneAddedAt: new Date() });
     }
-    const [account] = await models.Account.findOrCreate({ where });
+
+    if (email) {
+        account.update({ email, emailAddedAt: new Date() });
```
vgrichina: this is missing `await`

app.js
```diff
@@ -219,17 +302,20 @@ router.post('/account/sendRecoveryMessage', async ctx => {
...
     }
 
-    const where = { accountId };
+    const [account] = await models.Account.findOrCreate({ where: { accountId } });
+
     if (phoneNumber) {
-        where.phoneNumber = phoneNumber;
-    } else if (email) {
-        where.email = email;
+        account.update({ phoneNumber, phoneAddedAt: new Date() });
```
vgrichina: this is missing `await`

app.js
```diff
@@ -210,6 +282,17 @@ ${recoverUrl}
 
 const { parseSeedPhrase } = require('near-seed-phrase');
 
+router.post('/account/seedPhraseAdded', async ctx => {
+    const { accountId, securityCode, signature } = ctx.request.body;
+
+    await verifyAccountOwnership({ accountId, ctx, securityCode, signature });
+
+    const [account] = await models.Account.findOrCreate({ where: { accountId } });
+    account.update({ phraseAddedAt: new Date() });
```
vgrichina: this is missing `await`

app.js
```diff
@@ -147,6 +168,57 @@ router.post('/account/:phoneNumber/:accountId/validateCode', async ctx => {
...
+
+    if (recoveryMethod === 'phone') {
+        account.update({ phoneNumber: null, phoneAddedAt: null });
+    }
+    if (recoveryMethod === 'email') {
+        account.update({ email: null, emailAddedAt: null });
+    }
+    if (recoveryMethod === 'phrase') {
+        account.update({ phraseAddedAt: null });
+    }
```
vgrichina: these all are missing `await`

test/app.test.js
vgrichina: Likely caused by missing `await`, `.reload()` likely helps just by accident because it creates more delay.

test/app.test.js
```diff
@@ -77,4 +124,184 @@ describe('/account/sendRecoveryMessage', () => {
...
+        let response = await request.post('/account/deleteRecoveryMethod')
+            .send({
+                accountId,
+                recoveryMethod: 'phone',
+                signature,
+                securityCode,
+            });
+        expect(response.status).toBe(200);
+        await account.reload();
+        expect(response.body).toEqual(JSON.parse(JSON.stringify({
```
vgrichina: why is `JSON.parse(JSON.stringify` roundtrip needed?

app.js
```diff
@@ -34,8 +38,8 @@ const router = new Router();
 const creatorKeyJson = JSON.parse(process.env.ACCOUNT_CREATOR_KEY);
 const recoveryKeyJson = JSON.parse(process.env.ACCOUNT_RECOVERY_KEY);
 const keyStore = {
-    async getKey(networkId, accountId) {
-        if (accountId == creatorKeyJson.account_id) {
+    async getKey(accountId) {
```
vgrichina: Why this change? https://github.com/nearprotocol/nearlib/blob/de49029846a95790adb81ed977a0327517148692/src.ts/key_stores/keystore.ts#L10

app.js
```diff
@@ -147,6 +168,57 @@ router.post('/account/:phoneNumber/:accountId/validateCode', async ctx => {
...
+
+    ctx.body = recoveryMethodsFor(account || {});
+});
+
+const recoveryMethods = ['phone', 'email', 'phrase'];
+
+router.post('/account/deleteRecoveryMethod', async ctx => {
+    const { accountId, recoveryMethod, securityCode, signature } = ctx.request.body;
+
+    if (!recoveryMethods.includes(recoveryMethod)) {
```
vgrichina: feel like it's more clear to have `switch (recoveryMethod)`?  with all the cases (including `default:` for this one)

app.js
```diff
@@ -147,6 +168,57 @@ router.post('/account/:phoneNumber/:accountId/validateCode', async ctx => {
...
+        ctx.throw(400, `Given recoveryMethod '${recoveryMethod}' invalid; must be one of: ${recoveryMethods.join(', ')}`);
+    }
+
+    const account = await models.Account.findOne({ where: { accountId } });
+
+    if (!account) {
+        ctx.throw(404, `Could not find account with accountId: '${accountId}'`);
+    }
```
vgrichina: Should IMO live in middleware, similar to https://github.com/nearprotocol/near-fiddle-api/blob/c0ecec0b27fba1a0c961e7e0223c961ac113d50d/app.js#L59

test/app.test.js
chadoh: Casts the date objects created by `new Date()` to strings – can you think of a more descriptive way to do this?

chadoh: I've done this in a new way now; I like it more and I think you will too!

app.js
chadoh: I've updated it with something I think you'll like more

app.js
chadoh: I was able to use a switch statement based off of `e.status` here, and I like it much more. Thanks!

app.js
chadoh: I did this in its own commit, and at that time we only called `getKey` from one place in this codebase: https://github.com/nearprotocol/near-contract-helper/pull/88/commits/9805cdfb343d48c608ee8879c22446204a0bc285#diff-0364f57fbff2fabbe941ed20c328ef1aR116

```js
const helperPublicKey = (await keyStore.getKey(recoveryKeyJson.account_id)).publicKey;
```

We were only passing one argument to `getKey`, and that key appears to be an account id, not a network id. Maybe I misunderstand how this `keyStore` gets used, though?

Since that commit, I've added others which remove the call to `keyStore.getKey` altogether. Do we need to keep this `keyStore` definition around at all?

vgrichina: we can remove all stuff related to `recoveryKeyJson` given we aren't doing custodial recovery anymore.

`const helperPublicKey = (await keyStore.getKey(recoveryKeyJson.account_id)).publicKey;`

this was a typo, calling `getKey` in wrong way

app.js
```diff
@@ -108,43 +120,67 @@ router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
...
-        await account.update({ securityCode: null, confirmed: true });
-    } else {
-        await (await ctx.near.account(accountId)).addKey(publicKey);
-        await account.update({ securityCode: null });
+    ctx.throw(400, `Given recoveryMethod '${recoveryMethod}' invalid; must be one of: ${allRecoveryMethods.join(', ')}`);
+}
+
+async function checkAccountInDB(ctx, next) {
+    const { accountId } = ctx.request.body;
+    const account = await models.Account.findOne({ where: { accountId } });
```
vgrichina: any reason not to do it similarly to https://github.com/nearprotocol/near-fiddle-api/blob/c0ecec0b27fba1a0c961e7e0223c961ac113d50d/app.js#L60 so that `account` can be fetched from `ctx` instead of being queried from DB again?

app.js
```diff
@@ -108,43 +120,67 @@ router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
...
+        return;
     }
+    ctx.throw(404, `Could not find account with accountId: '${accountId}'`);
+}
 
-    ctx.body = {};
+router.post('/account/deleteRecoveryMethod', checkRecoveryMethod, checkAccountInDB, checkAccountOwnership, async ctx => {
+    const { accountId, recoveryMethod } = ctx.request.body;
+    const account = await models.Account.findOne({ where: { accountId } });
+    await recoveryMethodUpdaters[recoveryMethod](account);
```
vgrichina: It's not a common pattern (only used once) so I feel like this indirection is confusing. Maybe just use `switch`?

vgrichina: also somewhat confusing that these are "updaters", but actually they all delete info

vgrichina: if you want to make it generic – it makes sense to define separate `RecoveryMethod` or `Key` entity and then keep a list of these in `Account`

chadoh: I've renamed it to `recoveryMethodDeleters`, but I want to keep this pattern rather than using a switch statement. Here's what a switch statement would look like:

```js
const recoveryMethods = ['phone', 'email', 'phrase'];
async function checkRecoveryMethod(ctx, next) {
    const { recoveryMethod } = ctx.request.body;
    if (recoveryMethods.includes(recoveryMethod)) {
        await next();
        return;
    }
    ctx.throw(400, `Given recoveryMethod '${recoveryMethod}' invalid; must be one of: ${recoveryMethods.join(', ')}`);
}
router.post(
    '/account/deleteRecoveryMethod',
    checkRecoveryMethod,
    withAccount,
    checkAccountOwnership,
    async ctx => {
        switch (ctx.request.body.recoveryMethod) {
        case 'phone':
            await ctx.account.update({ phoneNumber: null, phoneAddedAt: null });
            break;
        case 'email':
            await ctx.account.update({ email: null, emailAddedAt: null });
            break;
        case 'phrase':
            await ctx.account.update({ phraseAddedAt: null });
        }
        ctx.body = recoveryMethodsFor(ctx.account);
    }
);
```

While more common, I don't like that this introduces some easy-to-miss coupling between `checkRecoveryMethod` and the body of the route – if we add a new recovery method, we will need to add an entry to `recoveryMethods` and adding a branch to the `switch` statement.

I sort of like the idea of adding this logic to the `Account` model, but right now the logic is small and contained right next to the place it's needed. Adding it to `Account` would clutter our global `Account` model and would, imo, result in harder-to-understand code than either option discussed so far.

test/app.test.js
```diff
@@ -16,7 +17,14 @@ process.env = {
 };
 const app = require('../app');
 
-const { KeyPair } = require('nearlib');
+const nearlib = require('nearlib');
+
+const NETWORK_ID = process.env.NETWORK_ID;
```
vgrichina: hmm, is setting `NETWORK_ID` now a requirement for `yarn test` to work? I'd like us to avoid it.

chadoh: I wasn't sure what to do about this. You'll see that `NETWORK_ID` is also being used in app.js, which got introduced at the same time that I removed old RECOVERY_KEY stuff and the custom keyStore: 3073a99d. That's when I decided it probably made sense to add it to the ENV.

How would you update this?

vgrichina: I think it generally isn't relevant in contract-helper context. It basically only matters when you have persistent key store (so that there can be different keys for different networks). In current contract-helper it would really work the same whether it is "default" or "bazinga" or "whatever", which means passing it externally is just extra hassle.

chadoh:
@vgrichina this is ready for re-review. You can test out the changes in NEAR Wallet by following the instructions in https://github.com/nearprotocol/near-wallet/pull/399

README.md
```diff
@@ -37,7 +37,7 @@ By default, it assumes that you're running a local node and local network. To do
 * clone [nearcore](https://github.com/nearprotocol/nearcore)
 * in your nearcore directory, get a local network running (at the time of writing, the command was `./scripts/start_localnet.py`)
 
-Note that you need to add a `ACCOUNT_CREATOR_KEY` and `ACCOUNT_RECOVERY_KEY` to your `.env`. Running `nearcore` locally created a `~/.near/validator_keys.json` file for you. Copy the contents of this file and paste them as a single line for each of the `*_KEY` values in your `.env`.
+Note that you need to add an `ACCOUNT_CREATOR_KEY` to your `.env`. Running `nearcore` locally created a `~/.near/validator_keys.json` file for you. Copy the contents of this file and paste them as a single line for the `*_KEY` value in your `.env`.
```
vgrichina: it's `validator_key.json` (without `s`)

app.js
```diff
@@ -18,33 +18,32 @@ app.use(async function(ctx, next) {
...
 const creatorKeyJson = JSON.parse(process.env.ACCOUNT_CREATOR_KEY);
-const recoveryKeyJson = JSON.parse(process.env.ACCOUNT_RECOVERY_KEY);
 const keyStore = {
-    async getKey(networkId, accountId) {
-        if (accountId == creatorKeyJson.account_id) {
-            return KeyPair.fromString(creatorKeyJson.secret_key || creatorKeyJson.private_key);
-        }
-        // For account recovery purposes use recovery key when updating any account
-        return KeyPair.fromString(recoveryKeyJson.secret_key || creatorKeyJson.private_key);
+    async getKey() {
```
vgrichina: I think at this point it makes sense to use stock `InMemoryKeyStore` and just add key from `creatorKeyJson` there. Custom logic was only needed for `recoveryKeyJson`

app.js
```diff
@@ -108,44 +131,71 @@ router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
...
 const bs58 = require('bs58');
-const verifySignature = async (nearAccount, securityCode, signature) => {
-    const hasher = crypto.createHash('sha256');
-    hasher.update(securityCode);
-    const hash = hasher.digest();
-    const helperPublicKey = (await keyStore.getKey(recoveryKeyJson.account_id)).publicKey;
-    const accessKeys = await nearAccount.getAccessKeys();
-    if (!accessKeys.find(it => it.public_key == helperPublicKey.toString())) {
-        throw Error(`Account ${nearAccount.accountId} doesn't have helper key`);
+const verifySignature = async (nearAccount, data, signedData) => {
```
vgrichina: Technically this is `signature` only, not `signedData` (same goes for `blockNumberSigned` – it's really `blockNumberSignature`). That's why `nacl.sign.detached` is used.  There is also variant which actually operates on signed data (https://github.com/dchest/tweetnacl-js/blob/master/README.md#naclsignopensignedmessage-publickey).  So I'm not sure why you decided to rename.

app.js
```diff
@@ -108,44 +131,71 @@ router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
...
+    try {
+        const hash = crypto.createHash('sha256').update(data).digest();
+        const accessKeys = await nearAccount.getAccessKeys();
+        return accessKeys.some(it => {
+            const publicKey = it.public_key.replace('ed25519:', '');
+            return nacl.sign.detached.verify(hash, Buffer.from(signedData, 'base64'), bs58.decode(publicKey));
+        });
+    } catch (e) {
+        console.error(e);
```
vgrichina: hmm, are there any expected errors? Shouldn't we just fail through normal failure path (i.e. 500) for unexpected errors?

app.js
```diff
@@ -108,44 +131,71 @@ router.post('/account/:phoneNumber/:accountId/requestCode', async ctx => {
...
-    const { phoneNumber, accountId } = ctx.params;
-    const { securityCode, signature, publicKey } = ctx.request.body;
+async function recoveryMethodsFor(account) {
+    if (!account) return [];
 
-    const account = await models.Account.findOne({ where: { accountId, phoneNumber } });
-    if (!account || !account.securityCode || account.securityCode != securityCode) {
-        ctx.throw(401);
+    return await account.getRecoveryMethods().map(recoveryMethod => ({
+        kind: recoveryMethod.kind,
```
vgrichina: nit: just FYI it's possible to filter included fields right on Sequelize level:
https://sequelize.org/master/class/lib/model.js~Model.html#static-method-findAll (provide `attributes` list same as for `findAll`.

not blocking anything, just letting you know it's a thing (as it's often somewhat faster)

app.js
```diff
@@ -219,18 +290,23 @@ router.post('/account/sendRecoveryMessage', async ctx => {
...
-    await sendRecoveryMessage({ ...account.dataValues, seedPhrase });
+    const [account] = await models.Account.findOrCreate({ where: { accountId } });
+    await account.createRecoveryMethod({
+        kind: phoneNumber ? 'phone' : 'email',
+        detail: phoneNumber || email,
+        publicKey,
+    });
+
+    await sendRecoveryMessage({
+        accountId: account.accountId,
```
vgrichina: nit: seems like can just be `accountId` here

migrations/20200304211821-unique-accountId.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+module.exports = {
+    up: (queryInterface) => {
+        return queryInterface.addIndex(
```
vgrichina: I think we might have problem with this migration as our current DB doesn't have unique account IDs. So we need some more complicated strategy (e.g. keep only latest entry for given accountId).

vgrichina: @chadoh I think it's best to take snapshot of current accounts DB from Render and test with it locally, to make sure all migrations apply cleanly.

migrations/20200325204238-mv-data-from-accounts-to-recoverymethods.js
```diff
@@ -0,0 +1,80 @@
...
+                        'SELECT * FROM "Accounts" WHERE "phoneNumber" IS NOT NULL',
+                        { transaction }
+                    ),
+                    queryInterface.sequelize.query(
+                        'SELECT * FROM "Accounts" WHERE "email" IS NOT NULL',
+                        { transaction }
+                    )
+                ]).then(([[phoneRecords], [emailRecords]]) => {
+                    Promise.all([
+                        phoneRecords.length && queryInterface.bulkInsert(
```
vgrichina: we have a small DB, so this might work (test with Render snapshot) – but generally it's best to run raw SQL for bulk ops like this vs loading all rows to RAM

migrations/20200325204238-mv-data-from-accounts-to-recoverymethods.js
```diff
@@ -0,0 +1,80 @@
...
+                    });
+                }).catch(err => {
+                    console.error(err);
+                    return transaction.rollback();
+                });
+            });
+        });
+    },
+
+    down: queryInterface => {
```
vgrichina: just FYI: I don't think there is any context where we currently use "down" migrations, so you might save time skipping them

chadoh: If we need to roll back a change we would need to roll back migrations, correct?

migrations/20200325204238-mv-data-from-accounts-to-recoverymethods.js
chadoh: I would prefer to do that but wasn't sure if Sequelize gave me the tools. I guess raw SQL would have been the way. I'll test this and see if it's acceptably performant.

migrations/20200304211821-unique-accountId.js
chadoh: Good call! Yes, I will test it out locally and modify this migration to work with testnet data before merging.

app.js
chadoh: Thanks for clarifying! I'll update the variable names.

app.js
chadoh: I tried that and though the tests passed locally, the app did not work in a deployed environment. I was getting an "unknown public key" error, iirc. See commit 22862aa for the `InMemoryKeyStore` implementation that I removed.

app.js
chadoh: If you provide a garbage signature, then the `verify` line throws an error. This was happening in the tests a fair bit at one point (not sure if it still would), and could potentially happen in the wild as well. If that happens, we want this function to return `false`. Do you agree?

migrations/20200325204238-mv-data-from-accounts-to-recoverymethods.js
vgrichina: in theory – yes. In practice I had to resort to manual rollbacks most of the time I was running Sequelize in production (few years). Down migrations don't get tested, state is inconsistent because of something unforeseen, etc.

migrations/20200325204238-mv-data-from-accounts-to-recoverymethods.js
vgrichina: given that this is one off, if it works on dump – no point in optimizing, it's more just for future cases

app.js
vgrichina: @chadoh I think we should better debug this (e.g. by logging what gets passed to `getKey` as arguments). Something else might be going wrong if wrong account is used to do stuff (that would explain public key not being found).

app.js
vgrichina: either one is ok from my POV. I'd probably just have let it throw out of laziness, but should be fine as is as well.

app.js
chadoh: Here's the version that didn't work:

```js
const creatorKeyJson = JSON.parse(process.env.ACCOUNT_CREATOR_KEY);
const keyStore = new nearlib.keyStores.InMemoryKeyStore();
const keyPair = nearlib.KeyPair.fromString(creatorKeyJson.secret_key || creatorKeyJson.private_key);
keyStore.setKey(process.env.NETWORK_ID, creatorKeyJson.account_id, keyPair);

const nearPromise = (async () => {
    const near = await nearlib.connect({
        deps: { keyStore },
        masterAccount: creatorKeyJson.account_id,
        nodeUrl: process.env.NODE_URL
    });
    return near;
})();
```

If I remember correctly, the error involved the `masterAccount: creatorKeyJson.account_id` line. It seemed like maybe the `keyPair = nearlib.KeyPair.fromString(...)` worked properly, but the `keyStore.setKey` didn't. I wondered if `keyStore.setKey` was running on app startup, but not re-running on every individual request to the app.

I doubt it's worth troubleshooting this right now. There's a lot of other stuff here that I want to land, and the current implementation works.

migrations/20200304211821-unique-accountId.js
chadoh: I added a new migration that runs before this one which deletes all but the most recent of each set of duplicates. Using yesterday's dump of production data, here's what running migrations looks like for me locally:

```
 contract-helper ⮀ recoveryMethods ⮀ psql accounts_development helper -c 'select count(*) from "Accounts";'
 count
-------
  2226
(1 row)

 contract-helper ⮀ recoveryMethods ⮀ sequelize db:migrate

Sequelize CLI [Node: 12.14.1, CLI: 5.5.1, ORM: 5.21.5]

Loaded configuration file "config/config.js".
Using environment "development".
== 20200304211820-drop-duplicate-accountIds: migrating =======
== 20200304211820-drop-duplicate-accountIds: migrated (0.067s)

== 20200304211821-unique-accountId: migrating =======
== 20200304211821-unique-accountId: migrated (0.045s)

== 20200325154731-add-recovery-methods: migrating =======
== 20200325154731-add-recovery-methods: migrated (0.016s)

== 20200325204048-constrain-recoverymethods: migrating =======
== 20200325204048-constrain-recoverymethods: migrated (0.043s)

== 20200325204238-mv-data-from-accounts-to-recoverymethods: migrating =======
== 20200325204238-mv-data-from-accounts-to-recoverymethods: migrated (0.231s)

== 20200325213728-drop-phone-and-email-from-accounts: migrating =======
== 20200325213728-drop-phone-and-email-from-accounts: migrated (0.008s)

 contract-helper ⮀ recoveryMethods ⮀ psql accounts_development helper -c 'select count(*) from "Accounts";'
 count
-------
  2138
(1 row)
```

PR #121, opened 2020-04-21
fix: allow null publicKey

chadoh:
Accounts created before we started saving publicKeys will not be able to send them along to these endpoints, so we need to allow them to be null

## to do

- [x] wait until #120 is merged, then retarget this PR at `master`

### DIFF
--- a/test/app.test.js
+++ b/test/app.test.js
@@ -314,4 +314,15 @@ describe('/account/deleteRecoveryMethod', () => {
         expect(response.body.length).toBe(1);
         expect(response.body.map(m => m.kind)).toEqual(['email']);
     });
+
+    test('does not return 400 for old accounts with publicKey=NULL', async () => {
+        const accountId = await createNearAccount();
+        const account = await models.Account.create({ accountId });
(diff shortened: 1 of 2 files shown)


test/app.test.js
```diff
@@ -314,4 +314,15 @@ describe('/account/deleteRecoveryMethod', () => {
...
+
+    test('does not return 400 for old accounts with publicKey=NULL', async () => {
+        const accountId = await createNearAccount();
+        const account = await models.Account.create({ accountId });
+        await account.createRecoveryMethod({ kind: 'phrase', publicKey: null });
+        const signature = await signatureFor(accountId);
+
+        let response = await request.post('/account/deleteRecoveryMethod')
+            .send({ accountId, recoveryMethod: 'phrase', publicKey: null, ...signature });
+        expect(response.status).toBe(200);
```
vgrichina, line 326: @chadoh shouldn't we also check that recovery method got deleted as well?

chadoh: Good point; test added here #123

PR #131, opened 2020-05-05
migrations for recovery methods

Patrick1904:
Migrations:
- Move securityCode from Account to RecoveryMethod
- Require RecoveryMethod to be unique (e.g. can only be one hi@near.org email)

### DIFF
--- a/migrations/20200429204207-move-security-code-to-recovery-methods.js
+++ b/migrations/20200429204207-move-security-code-to-recovery-methods.js
@@ -1,0 +1,31 @@
+'use strict';
+
+module.exports = {
+    up: (queryInterface, Sequelize) => {
+        return queryInterface.sequelize.transaction(transaction => {
+            return Promise.all([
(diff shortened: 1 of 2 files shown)


migrations/20200429204207-move-security-code-to-recovery-methods.js
```diff
@@ -0,0 +1,31 @@
...
+                    type: Sequelize.STRING
+                }, {transaction}),
+                queryInterface.removeColumn('Accounts', 'securityCode', {transaction}),
+                queryInterface.removeColumn('Accounts', 'confirmed', {transaction}),
+            ]);
+        });
+    },
+
+    down: (queryInterface, Sequelize) => {
+        return queryInterface.sequelize.transaction(transaction => {
```
vgrichina, line 17: @Patrick1904 tip (no action required): we don't really use `down` migrations anywhere in our process.

PR #132, opened 2020-05-05
add recovery method verification

Patrick1904:
[Wallet changes (559)](https://github.com/near/near-wallet/pull/559)
[Migrations and model changes for this code (135)](https://github.com/near/near-contract-helper/pull/135)

Covered in this PR:
- Initialize an ‘unconfirmed’ recovery method (email/phone) without a publicKey and adding a securityCode
- Validate securityCode
- Add publicKey to Recoverymethod (‘confirmed’) before sending recovery message
- Create new endpoint to ‘re-send’ recovery link

### DIFF
--- a/app.js
+++ b/app.js
@@ -187,7 +187,7 @@ router.post(
     checkAccountOwnership,
     async ctx => {
         const { accountId, seedPhrase, publicKey, method } = ctx.request.body;
-        const [account] = await models.Account.findOrCreate({ where: { accountId } });
+        const account = await models.Account.findOne({ where: { accountId } });
 
         const [recoveryMethod] = await account.getRecoveryMethods({ where: {
             kind: method.kind,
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -149,8 +129,13 @@ async function recoveryMethodsFor(account) {
     if (!account) return [];
 
     return await account.getRecoveryMethods({
-        attributes: ['createdAt', 'detail', 'kind', 'publicKey']
-    }).map(m => m.toJSON());
+        attributes: ['createdAt', 'detail', 'kind', 'publicKey', 'securityCode']
+    }).map(method => {
+        const json = method.toJSON();
+        json.confirmed = !method.securityCode;
```
vgrichina, line 135: best to define virtual field for this https://sequelize.org/master/manual/getters-setters-virtuals.html#virtual-fields

app.js
```diff
@@ -149,8 +129,13 @@ async function recoveryMethodsFor(account) {
     if (!account) return [];
 
     return await account.getRecoveryMethods({
-        attributes: ['createdAt', 'detail', 'kind', 'publicKey']
-    }).map(m => m.toJSON());
+        attributes: ['createdAt', 'detail', 'kind', 'publicKey', 'securityCode']
+    }).map(method => {
+        const json = method.toJSON();
```
vgrichina, line 134: I think `.toJSON()` will be done automatically (i.e. no need for `.map`)

app.js
```diff
@@ -198,6 +183,40 @@ router.post(
...
 
+router.post(
+    '/account/resendRecoveryLink',
+    checkAccountOwnership,
+    async ctx => {
+        const { accountId, seedPhrase, publicKey, method } = ctx.request.body;
+        const [account] = await models.Account.findOrCreate({ where: { accountId } });
+        let email, phoneNumber;
+
+        if (method.kind === 'email') {
```
vgrichina, line 194: nit: seems like we can just make `sendRecoveryMessage` accept `method` directly instead of `phoneNumber` / `email`

app.js
```diff
@@ -198,6 +183,40 @@ router.post(
...
+            email = method.detail;
+        } else if (method.kind === 'phone') {
+            phoneNumber = method.detail;
+        }
+
+        const where = {
```
vgrichina, line 200: nit: seems cleaner to just inline this with `getRecoveryMethods` call

app.js
```diff
@@ -282,11 +301,76 @@ router.post(
...
+        ctx.body = await recoveryMethodsFor(account);
+    }
+);
+
+router.post('/account/validateSecurityCode',
+    checkAccountOwnership,
+    async ctx => {
+        const { accountId, phoneNumber, email, securityCode } = ctx.request.body;
+
+        const [account] = await models.Account.findOrCreate({ where: { accountId } });
```
vgrichina, line 353: hmm, shouldn't account already exist by the time we need to validate security code? Looks like shouldn't be using `findOrCreate`

app.js
```diff
@@ -282,11 +301,76 @@ router.post(
...
+
+        const where = {
+            kind: phoneNumber ? 'phone' : 'email',
```
vgrichina, line 356: why not pass `kind` and `detail` (i.e. `method`) directly as in `initializeRecoveryMethod`?

PR #152, opened 2020-05-26
2FA Refactor

mattlockyer:
Eager to make this repo more modular.
This PR refactors a number of methods previously used by routes defined in app.js into their own modules.
Changes to project structure:
```
/providers
  sms.js
  email.js
/routes
  account.js
  recovery.js
  2fa.js (wip)
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -187,7 +187,7 @@ router.post(
     checkAccountOwnership,
     async ctx => {
         const { accountId, seedPhrase, publicKey, method } = ctx.request.body;
-        const [account] = await models.Account.findOrCreate({ where: { accountId } });
+        const account = await models.Account.findOne({ where: { accountId } });
 
         const [recoveryMethod] = await account.getRecoveryMethods({ where: {
             kind: method.kind,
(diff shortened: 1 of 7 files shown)


mattlockyer:
Tested locally, but not certain if this coverage would be affected by the refactor.

Flags: worried about missing references e.g. a method in runtime not having access to the same variable scope they had when in `app.js`

image

app.js
```diff
@@ -54,221 +54,81 @@ app.use(async (ctx, next) => {
...
-            where: {
-                kind: ctx.request.body.recoveryMethod,
-                publicKey: ctx.publicKey,
-            }
-        });
+        const { kind, publicKey } = ctx.request.body;
+        const [recoveryMethod] = await ctx.account.getRecoveryMethods({ where: {
+            kind: kind,
+            publicKey: publicKey,
+        }});
```
chadoh, line 102: I would prefer that style changes like this come from a linter and get added all at once; this PR is already going to be pretty big without a bunch of style-change noise

vgrichina:
@mattlockyer overall comments:

- feel like `providers` should just be `utils`, they aren't really providing any objects or smth like that
- what is put in `routes` mostly looks like middleware (`withAccount`, etc) vs specific route handlers
- needs rebase on `master`
- best way to test is actually use it with near-wallet. Unit tests are nice for smoke check and regression testing, but ultimate test is whether this works well with wallet.

vgrichina:
@mattlockyer should we close this one?

PR #165, opened 2020-06-17
Clean refactor pre-2fa

mattlockyer:
Moves some methods required by 2fa out of app.js and into
- utils/sms.js
- utils/email.js
- utils/near.js

In the second commit, these functions are removes from app.js and imported instead.

The is a require refactor pre-2fa since 2fa depends on these utils as well.

### DIFF
--- a/app.js
+++ b/app.js
@@ -325,7 +325,7 @@ const sendSecurityCode = async (securityCode, method) => {
 router.post('/account/initializeRecoveryMethod',
     checkAccountOwnership,
     async ctx => {
-        const { accountId, method, testing } = ctx.request.body;
+        const { accountId, method} = ctx.request.body;
         const [account] = await models.Account.findOrCreate({ where: { accountId } });
 
         let [recoveryMethod] = await account.getRecoveryMethods({ where: {
@@ -343,10 +343,6 @@ router.post('/account/initializeRecoveryMethod',
         const securityCode = password.randomPassword({ length: SECURITY_CODE_DIGITS, characters: password.digits });
         await recoveryMethod.update({ securityCode });
         await sendSecurityCode(securityCode, method);
-
-        if (testing) {
-            return ctx.body = securityCode;
-        }
 
         ctx.body = await recoveryMethodsFor(account);
     }
--- a/app.js
+++ b/app.js
@@ -1,4 +1,3 @@
-const nearAPI = require('near-api-js');
 const Koa = require('koa');
 const app = new Koa();
 
@@ -38,7 +37,7 @@ const router = new Router();
 const {
     creatorKeyJson,
     nearPromise,
-} = require('./utils/near')
+} = require('./utils/near');
 
 app.use(async (ctx, next) => {
     ctx.near = await nearPromise;
@@ -85,7 +84,7 @@ const password = require('secure-random-password');
 const models = require('./models');
 const SECURITY_CODE_DIGITS = 6;
 
-const sendSms = require('./utils/sms')
+const sendSms = require('./utils/sms');
 
 const nacl = require('tweetnacl');
 const crypto = require('crypto');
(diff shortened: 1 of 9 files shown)


mattlockyer:
Looks like checks have passed. Is there anything else blocking?

This is the minimum refactor of shared methods needed to be broken out before I can add more 2fa PRs.

utils/near.js
```diff
@@ -0,0 +1,51 @@
...
+const nearPromise = (async () => {
+    const near = await nearAPI.connect({
+        deps: { keyStore },
+        masterAccount: creatorKeyJson && creatorKeyJson.account_id,
+        nodeUrl: process.env.NODE_URL
+    });
+    return near;
+})();
+
+const getContract = async (contractName, viewMethods, changeMethods, secretKey) => {
```
vgrichina, line 27: where is `getContract` called?

app.js
```diff
@@ -35,28 +34,11 @@ app.use(async function(ctx, next) {
...
-        return null;
-    }
-})();
-const keyStore = {
-    async getKey() {
-        return nearAPI.KeyPair.fromString(creatorKeyJson.secret_key || creatorKeyJson.private_key);
-    }
-};
+const {
+    creatorKeyJson,
```
vgrichina, line 38: Makes sense to extract all the code using `creatorKeyJson` into module and just export middleware handling account creation. Any reason not to do it?

app.js
```diff
@@ -35,28 +34,11 @@ app.use(async function(ctx, next) {
...
+    nearPromise,
```
vgrichina, line 39: `nearPromise` also used only in one middleware, makes sense to export middleware `withNear` instead

utils/near.js
mattlockyer: In an upcoming PR. For now it's just a helper method, shouldn't really be an issue to include it here.

app.js
mattlockyer: Seems like a majority of the near code should go in `middleware/near.js` vs. `utils/near.js` correct?

And we can tighten this: https://github.com/near/near-contract-helper/blob/d70186d6f37625eedf238e048c2a2042e143745f/app.js#L37
Into a proper middleware `withNear` as you mentioned in: https://github.com/near/near-contract-helper/pull/165/files/d70186d6f37625eedf238e048c2a2042e143745f#r445233017

app.js
mattlockyer: Answered in https://github.com/near/near-contract-helper/pull/165/files/d70186d6f37625eedf238e048c2a2042e143745f#r445239440

utils/near.js
vgrichina: Let's include it when it's actually used. E.g. it probably shouldn't use `creatorKeyJson` at all. 

Same actually might apply to `withNear` as well. Looks like we ideally should depend on `creatorKeyJson` only in middleware which is using it to create accounts.

utils/near.js
```diff
@@ -0,0 +1,51 @@
...
+        async getKey() {
+            return nearAPI.KeyPair.fromString(secretKey);
+        },
+    };
+    const near = await nearAPI.connect({
+        deps: { keyStore },
+        masterAccount: creatorKeyJson && creatorKeyJson.account_id,
+        nodeUrl: process.env.NODE_URL
+    });
+    const contractAccount = new nearAPI.Account(near.connection, contractName);
```
vgrichina, line 38:
```js
const contractAccount = await near.account(contractName);
```

app.js
vgrichina: yeah, seems good to me

mattlockyer:
Closing to trigger a new build on Render with a new PR for the same branch.

mattlockyer:
New PR here to trigger Render build: https://github.com/near/near-contract-helper/pull/188

PR #188, opened 2020-07-16
Refactor Pre-2FA

mattlockyer:
Previous PR:
https://github.com/near/near-contract-helper/pull/165

This PR triggers render. It should be tested, rebased and then merged into master prior to adding 2FA features for v0 of wallet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea
 node_modules/
 .env
+yarn.lock
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 .idea
 node_modules/
 .env
-yarn.lock
--- a/app.js
+++ b/app.js
@@ -325,7 +325,7 @@ const sendSecurityCode = async (securityCode, method) => {
 router.post('/account/initializeRecoveryMethod',
     checkAccountOwnership,
     async ctx => {
-        const { accountId, method, testing } = ctx.request.body;
+        const { accountId, method} = ctx.request.body;
         const [account] = await models.Account.findOrCreate({ where: { accountId } });
 
         let [recoveryMethod] = await account.getRecoveryMethods({ where: {
@@ -343,10 +343,6 @@ router.post('/account/initializeRecoveryMethod',
         const securityCode = password.randomPassword({ length: SECURITY_CODE_DIGITS, characters: password.digits });
         await recoveryMethod.update({ securityCode });
         await sendSecurityCode(securityCode, method);
-
-        if (testing) {
-            return ctx.body = securityCode;
-        }
 
         ctx.body = await recoveryMethodsFor(account);
     }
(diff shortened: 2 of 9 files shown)


vgrichina:
@mattlockyer looks like tests are failing on CI

mattlockyer:
@vgrichina I don't know why they fail on Jenkins, looks like timeouts, but I can run them locally.

There is 1 test that I expect to fail, because we no longer require that the signature comes from a key that's a full access key (if the user has enabled 2fa).

I changed the filter condition on verifySignature (now in middleware/near.js) so we accept signatures from limited access keys that still recover to the accountId. What do you think about this?

Should I remove the test?

image

vgrichina:
> There is 1 test that I expect to fail, because we no longer require that the signature comes from a key that's a full access key (if the user has enabled 2fa).

@mattlockyer 

1) let's make sure tests actually pass on CI for every PR. If test fails it is necessary to either fix code or test (if something is wrong about it), otherwise PR won't be ready to merge.
2) If this is just "refactor" then by definition none of the tests should be failing. If there are intended logic changes – they need to be described in the description of the PR and tests need to be updated accordingly.

mattlockyer:
> There is 1 test that I expect to fail, because we no longer require that the signature comes from full access key (if the user has enabled 2fa).
> I changed the filter condition on verifySignature (now in middleware/near.js) so we accept signatures from limited access keys that still recover to the accountId. What do you think about this?
> Should I remove the test?

Do you see any security issue if verifySignature, which is used by checkAccountOwnership is not checking if the key is full access?

https://github.com/near/near-contract-helper/blob/fab413eb6be71cd9c0468a107565c9172f744f3b/middleware/near.js#L11-L22

mattlockyer:
> 2. If this is just "refactor" then by definition none of the tests should be failing. If there are intended logic changes – they need to be described in the description of the PR and tests need to be updated accordingly.

I see your point. Ok if we want to merge this I'll roll back the filter change and move to the 2fa branch.

Will try to fix tests now.

mattlockyer:
I'll try to replicate Travis CI locally.

My tests pass now that I rolled back the verifySignature change:

image

vgrichina:
> Do you see any security issue if verifySignature, which is used by checkAccountOwnership is not checking if the key is full access?

yes, this would allow any app which has login with NEAR to access near-contract-helper. `permission.FunctionCall.method_names.includes('__wallet__metadata'))` this was checking for special access key that is used for authorizing access when using Ledger. 
I think we need to use similar approach for 2FA.

mattlockyer:
Thanks for the help! I will check that it's a multisig key by checking the contract methods, but I'll push that in the 2fa PR.

mattlockyer:
@vgrichina 
check the test change here please, I think it's correct, was not passing before (locally, but not in CI)
https://github.com/near/near-contract-helper/pull/188/commits/6391f47bd75d7cdd24c32246cab96697ed4e7c8a

mattlockyer:
Now I just seem to be getting the downstream timeouts causing errors to fail.

When errors fail, they fail with 400, but none of our other codes.

And all tests pass locally.

PR #189, opened 2020-07-16
2FA Phase 0 support for wallet

mattlockyer:
This PR adds support for initializing and supporting 2FA accounts from wallet.
Routes are added in app.js
Builds on this PR: https://github.com/near/near-contract-helper/pull/188
Changes are mostly additions in `middleware/2fa.js`


mattlockyer:
@vgrichina cleaning some things up here:

1. looks like everything is passing so no existing functionality will be broken.
2. the PR test failing is a lock file thing, I deleted and ignored them, I think I mistakenly used npm at some point

Should I add tests for 2fa to `app.test.js` so we know what to expect going forward?
Maybe we should start a new file? `twoFactor.test.js`?

.gitignore
```diff
@@ -1,3 +1,5 @@
 .idea
 node_modules/
 .env
+yarn.lock
```
vgrichina, line 4: shouldn't ignore yarn.lock

app.js
```diff
@@ -74,6 +91,9 @@ async function recoveryMethodsFor(account) {
     });
 }
 
+/********************************
```
vgrichina, line 94: seems like need a rebase after my changes here https://github.com/near/near-contract-helper/pull/188

vgrichina:
> Should I add tests for 2fa to app.test.js so we know what to expect going forward?
> Maybe we should start a new file? twoFactor.test.js?

yeah, I think it's good to have a new file for 2fa tests.

app.js
mattlockyer: I keep getting into a rebase loop here, maybe you can help me or have a different strategy.

mattlockyer:
@vgrichina looks like tests are in good working order now. Everything we discussed yesterday is in:

1. yarn.lock is back
2. package.json is latest from master
3. new app tests - init and verify temp accounts, race conditions, clear all recovery methods and keep only the one we send message to (if new account) when sending recovery message
4. 2fa tests - everything except multisig add_request and confirm on chain

mattlockyer:
@vgrichina when you have a second can you check the latest changes to code expiry and the tests for 2fa?

Here's the code expiry commit: https://github.com/near/near-contract-helper/pull/189/commits/76125d85a0abe4aa12446c1f0c597ccc2e756349

I have updated some 2fa messaging as well. I still need to add messaging around adding a FAK to your account if 2FA is enabled but this shouldn't block tomorrow's release.

app.js
```diff
@@ -3,6 +3,7 @@ const app = new Koa();
 
 const body = require('koa-json-body');
 const cors = require('@koa/cors');
+// const { Op } = require('sequelize');
```
vgrichina, line 6: nit: remove altogether?

app.js
```diff
@@ -19,13 +20,14 @@ app.use(async function(ctx, next) {
...
         case 400:
+        case 401:
         case 403:
         case 404:
             ctx.throw(e);
             break;
         default:
             // TODO: Figure out which errors should be exposed to user
-            console.error('Error: ', e, JSON.stringify(e));
+            // console.error('Error: ', e, JSON.stringify(e));
```
vgrichina, line 30: why comment out?

app.js
```diff
@@ -267,6 +291,7 @@ const completeRecoveryValidation = async ctx => {
     if (!recoveryMethod) {
         ctx.throw(401);
     }
+    console.log(securityCode);
```
vgrichina, line 294: leftover debug log?

app.js
```diff
@@ -287,12 +317,26 @@ router.post('/account/sendRecoveryMessage', async ctx => {
...
 
     const account = await models.Account.findOne({ where: { accountId } });
-    const [recoveryMethod] = await account.getRecoveryMethods({ where: {
-        kind: method.kind,
-        detail: method.detail
-    }});
 
-    await recoveryMethod.update({ publicKey, securityCode: null });
+    if (isNew) {
+        // remove all recovery methods
```
vgrichina, line 322: nit: makes sense to explain that these might be coming from other users, etc

app.js
```diff
@@ -287,12 +317,26 @@ router.post('/account/sendRecoveryMessage', async ctx => {
...
+        const allRecoveryMethods = await account.getRecoveryMethods();
+        for (const rm of allRecoveryMethods) {
+            await rm.destroy();
+        }
+        // set up the official recovery method for this account
```
vgrichina, line 327: what do you mean by "official" method?

app.js
```diff
@@ -287,12 +317,26 @@ router.post('/account/sendRecoveryMessage', async ctx => {
...
+        await account.createRecoveryMethod({
```
vgrichina, line 328: isn't `RecoveryMethod` instance already created in `/account/initializeRecoveryMethodForTempAccount`?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+const crypto = require('crypto');
+const nacl = require('tweetnacl');
+const { creatorKeyJson } = require('./near');
+const { sendSms } = require('../utils/sms');
+const { sendMail } = require('../utils/email');
+const models = require('../models');
+const Sequelize = require('sequelize');
+const Op = Sequelize.Op;
+const password = require('secure-random-password');
+const SECURITY_CODE_DIGITS = 6;
```
vgrichina, line 11: nit: let's have some space between imports and constants

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+
+const getContract = async (contractName, secretKey) => {
+    const keyStore = {
+        async getKey() {
+            return nearAPI.KeyPair.fromString(secretKey);
+        },
+    };
+    const near = await nearAPI.connect({
+        deps: { keyStore },
+        masterAccount: creatorKeyJson && creatorKeyJson.account_id,
```
vgrichina, line 24: no need for `masterMaster` account and `creatorKeyJson` here as we shouldn't be creating any accounts here

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    const res = await contract.confirm({ request_id: parseInt(request_id) }).catch((e) => {
+        return { success: false, error: e };
+    });
+    return { success: !!res, res };
+};
+/********************************
+2FA Key
+********************************/
+// helper: generates a deterministic key based on the accountId
+const getDetermKey = async (accountId) => {
```
vgrichina, line 47: why not do this in `getKey` of `keyStore`?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        viewMethods,
+        changeMethods,
+    });
+    return contract;
+};
+const confirmRequest = async (accountId, request_id) => {
+    const key = await getDetermKey(accountId);
+    const contract = await getContract(accountId, key.secretKey);
+    // always parseInt on requestId. requestId is string in db because it could come from url params etc...
+    const res = await contract.confirm({ request_id: parseInt(request_id) }).catch((e) => {
```
vgrichina, line 38: best to specify second argument to `parseInt` if you don't want surprises

vgrichina, line 38: why use `.catch` here instead of using `try` / `catch` block?

Using `.then` / `.catch` in `async` function is confusing. I think there is already mistake here:

`return { success: false, error: e };` will effectively return result of the promise to `const res`. So `return { success: !!res, res };` will return `{ success: true, { success: false, error: e}}`. Which suggests that error condition wasn't tested here.

I think it's generally best to stick to `try` / `catch` in `async` functions and let most exceptions fall through until route middleware. Like there is no reason to return error as successful HTTP response with `success: false` when appropriate HTTP error can be returned instead (e.g. 401 in this case).

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        return { success: false, error: e };
+    });
+    return { success: !!res, res };
+};
+/********************************
+2FA Key
+********************************/
+// helper: generates a deterministic key based on the accountId
+const getDetermKey = async (accountId) => {
+    const hash = crypto.createHash('sha256').update(accountId + creatorKeyJson.private_key).digest();
```
vgrichina, line 48: let's use separate seed for this than creator key

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    const keyPair = nacl.sign.keyPair.fromSeed(hash);//nacl.sign.keyPair.fromSecretKey(hash)
```
vgrichina, line 49: nit: remove comment

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    return {
+        publicKey: `ed25519:${nearAPI.utils.serialize.base_encode(keyPair.publicKey)}`,
```
vgrichina, line 51: nit: import `base_encode` directly so that no need to repeat this whole thing?

```
const { utils: { serialize: { base_encode } } } = require('near-api-js');
```

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        secretKey: nearAPI.utils.serialize.base_encode(keyPair.secretKey)
+    };
+};
+/********************************
+Sending Codes
+method.kind = ['2fa-email', '2fa-phone']
+********************************/
+const prettyRequestInfo = ({ request }) => `
+    Transaction Recipient: ${ request.receiver_id }
+    Actions:\n\t${ request.actions.map((r) => r.type + (r.amount ? ': ' + nearAPI.utils.format.formatNearAmount(r.amount, 4) : '')).join('\n\t') }
```
vgrichina, line 61: We need to have user-friendly formatting for actions.

@kcole16 may you help with copy for different type of tx actions? Should we basically match what wallet / explorer do?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+            <td align="center">
+                <img src="https://near.org/wp-content/themes/near-19/assets/downloads/near_logo.png" width="300" height="117"
+            </td>
+        </tr>
+        <tr>
+            <td>
+                <p>NEAR Wallet security code: ${securityCode}</p>
+                <p><strong>Important:</strong> By entering this code, you are authorizing the following transaction:\n\n</p>
+                <pre>
+                    ${dataOutput}
```
vgrichina, line 88: let's make sure HTML cannot be injected here (e.g. by including HTML into function call data)

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+Checking code_hash (is multisig deployed)
+********************************/
+const isContractDeployed = async(accountId) => {
+    const keyStore = {
+        async getKey() {
+            return nearAPI.KeyPair.fromString('bs');
+        },
+    };
+    // check account code_hash
+    const near = await nearAPI.connect({
```
vgrichina, line 121: why do we need different `near` instance than for everything else?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        deps: { keyStore },
+        masterAccount: creatorKeyJson && creatorKeyJson.account_id,
+        nodeUrl: process.env.NODE_URL
+    });
+    const nearAccount = new nearAPI.Account(near.connection, accountId);
+    const state = await nearAccount.state();
+    console.log(state);
+    return state.code_hash !== '11111111111111111111111111111111';
```
vgrichina, line 129: Need to specify desired code hash as environment variable. Definitely shouldn't treat all accounts with contract deployed as multisig contract.

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+};
+
+/********************************
+@warn protect these routes using checkAccountOwnership middleware from app.js
+@warn Requires refactor
+********************************/
+// http post http://localhost:3000/2fa/getAccessKey accountId=mattlock
```
vgrichina, line 136: let's just keep localhost example?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+// http post https://near-contract-helper-2fa.onrender.com/2fa/getAccessKey accountId=mattlock
+// Call this to get the public key of the access key that contract-helper will be using to confirm multisig requests
+const getAccessKey = async (ctx) => {
+    const { accountId } = ctx.request.body;
+    const { publicKey } = await getDetermKey(accountId);
+    ctx.body = {
+        success: true,
+        publicKey
+    };
+};
```
vgrichina, line 147: nit: empty line to separate these blocks?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+********************************/
+// http post http://localhost:3000/2fa/getAccessKey accountId=mattlock
+// http post https://helper.testnet.near.org/2fa/getAccessKey accountId=mattlock
+// http post https://near-contract-helper-2fa.onrender.com/2fa/getAccessKey accountId=mattlock
+// Call this to get the public key of the access key that contract-helper will be using to confirm multisig requests
+const getAccessKey = async (ctx) => {
+    const { accountId } = ctx.request.body;
+    const { publicKey } = await getDetermKey(accountId);
+    ctx.body = {
+        success: true,
```
vgrichina, line 144: why is `success: true` needed? HTTP already has success / non-success statuses: https://httpstatuses.com

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    }
+    const hasContractDeployed = await isContractDeployed(accountId);
+    // check recover methods
+    let [twoFactorMethod] = await account.getRecoveryMethods({ where: {
+        kind: {
+            [Op.startsWith]: '2fa-'
+        },
+    }});
+    // check if multisig contract is already deployed
+    if (hasContractDeployed || !!code_hash) {
```
vgrichina, line 178: not sure how `code_hash` is used?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        return;
+    }
+    const hasContractDeployed = await isContractDeployed(accountId);
+    // check recover methods
+    let [twoFactorMethod] = await account.getRecoveryMethods({ where: {
+        kind: {
+            [Op.startsWith]: '2fa-'
+        },
+    }});
+    // check if multisig contract is already deployed
```
vgrichina, line 177: why use comment to mark section vs extract into separate method / middleware (e.g. `checkContractIsDeployed`)

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    if (hasContractDeployed || !!code_hash) {
+        // check to see if they already have at least 1 twoFactorMethod
+        if (twoFactorMethod) {
+            ctx.throw(401, 'account with multisig contract already has 2fa method');
```
vgrichina, line 181: hmm, so does it mean that 2FA method cannot be changed later?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        } else {
+            // unlikely
+            twoFactorMethod = await account.createRecoveryMethod({ kind, detail, requestId: -1 });
+        }
+    } else {
+        // as long as the multisig is not deployed, can keep updating (or create new, 2fa method)
+        if (!twoFactorMethod) {
+            twoFactorMethod = await account.createRecoveryMethod({ kind, detail, requestId: -1 });
```
vgrichina, line 189: seems like this is happening in both of branches, so makes sense to check `twoFactorMethod` before `hasContractDeployed` unless I'm missing something

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        }
+    }
+    // check if 2fa method matches existing recovery method
+    const [recoveryMethod] = await account.getRecoveryMethods({ where: {
+        kind: kind.split('2fa-')[1],
+        detail,
+    }});
+    if (!recoveryMethod) {
+        // client waits to deploy contract until code is verified
+        const securityCode = await sendCode(method, twoFactorMethod);
```
vgrichina, line 205: hmm, why this requires both `method` and `twoFactorMethod`?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        const body = {
+            success: true,
```
vgrichina, line 207: see above about `success: true`

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+            message: '2fa initialized and code sent to verify method',
```
vgrichina, line 208: why is message needed? If there is different logic on wallet side, then should use response code like `codeSent` instead of message.

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        };
+        if (testing) {
```
vgrichina, line 210: So do I understand correctly that now I can just pass `testing=true` in the request to avoid need to supply valid security code? I don't think we should allow that.

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+            body.securityCode = securityCode;
+        }
+        ctx.body = body;
+    } else {
```
vgrichina, line 214: nit: I think makes sense to use early return for this to avoid unnecessary nesting https://www.itamarweiss.com/personal/2018/02/28/return-early-pattern.html

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    }
+};
+// http post http://localhost:3000/2fa/send accountId=mattlock method:='{"kind":"2fa-email","detail":"matt@near.org"}'
+// http post https://near-contract-helper-2fa.onrender.com/2fa/send accountId=mattlock method:='{"kind":"2fa-email","detail":"matt@near.org"}'
+// Call anytime after calling initCode to resend a new code, the new code will overwrite the old code
+const sendNewCode = async (ctx) => {
+    const { accountId, method, requestId, data } = ctx.request.body;
+    const [account] = await models.Account.findOrCreate({ where: { accountId } });
+    if (!account) {
+        console.warn('account should be created');
```
vgrichina, line 228: this is going to show up in server logs. If we want it to be useful – need to include more context like account name.

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        ctx.throw(401);
+    }
+    let [twoFactorMethod] = await account.getRecoveryMethods({ where: {
+        kind: {
+            [Op.startsWith]: '2fa-'
+        },
+    }});
+    if (!twoFactorMethod) {
+        console.warn('2fa not enabled');
```
vgrichina, line 237: see above about logs

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+// http post https://near-contract-helper-2fa.onrender.com/2fa/send accountId=mattlock method:='{"kind":"2fa-email","detail":"matt@near.org"}'
+// Call anytime after calling initCode to resend a new code, the new code will overwrite the old code
+const sendNewCode = async (ctx) => {
+    const { accountId, method, requestId, data } = ctx.request.body;
+    const [account] = await models.Account.findOrCreate({ where: { accountId } });
+    if (!account) {
+        console.warn('account should be created');
+        ctx.throw(401);
+    }
+    let [twoFactorMethod] = await account.getRecoveryMethods({ where: {
```
vgrichina, line 231: Seems like makes sense to do this in helper method (given that also done above and below).

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    const [twoFactorMethod] = await account.getRecoveryMethods({ where: {
+        securityCode,
+        kind: {
+            [Op.startsWith]: '2fa-'
+        },
+        // only verify codes that are 5 minutes old (if testing make this impossible)
+        updatedAt: {
+            [Op.gt]: Date.now() - (!testing ? 300000 : -1000)
+        }
+        // cannot test for requestId equality with negative integer???
```
vgrichina, line 261: hmm?

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    const { accountId, securityCode, requestId, testing } = ctx.request.body;
+    const account = await models.Account.findOne({ where: { accountId } });
+    const [twoFactorMethod] = await account.getRecoveryMethods({ where: {
+        securityCode,
+        kind: {
+            [Op.startsWith]: '2fa-'
+        },
+        // only verify codes that are 5 minutes old (if testing make this impossible)
+        updatedAt: {
+            [Op.gt]: Date.now() - (!testing ? 300000 : -1000)
```
vgrichina, line 259: probably best to do this check in JS after retrieving `twoFactorMethod` (going to be less confusing)

vgrichina, line 259: Let's not EVER have `testing` flags passed as parameter from external request. It's super easy way to get security holes. 

Best to restructure tests to not need it, but if absolutely necessary – should be possible to use environment variable, e.g. `process.env.NODE_ENV == 'test'`.

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+        }
+        // cannot test for requestId equality with negative integer???
+    }});
+    // checking requestId here with weak equality (no type match)
```
vgrichina, line 263: let's not have  comments which literally describe the same thing as next line of code. `twoFactorMethod.requestId != requestId` already says same thing. More interesting generally is why is code doing smth.

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    if (!twoFactorMethod || twoFactorMethod.requestId != requestId) {
+        console.warn('2fa code invalid');
+        ctx.throw(401);
+    }
+    //security code was valid, remove it and if there was a request, confirm it, otherwise just return success: true
```
vgrichina, line 268: see above about comments

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    await twoFactorMethod.update({ requestId: -1, securityCode: null });
+    // requestId already matched twoFactorMethod record and if it's not -1 we will attempt to confirm the request
+    if (requestId !== -1) {
+        ctx.body = await confirmRequest(accountId, parseInt(requestId));
+    } else {
```
vgrichina, line 273: see above about early return pattern

middleware/2fa.js
```diff
@@ -0,0 +1,287 @@
...
+    };
+};
+// http post http://localhost:3000/2fa/verify accountId=mattlock securityCode=430888
+// http post https://near-contract-helper-2fa.onrender.com/2fa/verify accountId=mattlock securityCode=437543
+// call when you want to verify the "current" securityCode
+const verifyCode = async (ctx) => {
+    const { accountId, securityCode, requestId, testing } = ctx.request.body;
+    const account = await models.Account.findOne({ where: { accountId } });
+    const [twoFactorMethod] = await account.getRecoveryMethods({ where: {
+        securityCode,
```
vgrichina, line 253: looks like if I don't pass `securityCode` (it's undefined) – I'm in?

middleware/near.js
```diff
@@ -44,6 +55,25 @@ async function checkAccountOwnership(ctx, next) {
     return await next();
 }
 
+
+function getAccount(ctx, accountId) {
+    return new nearAPI.Account(ctx.near.connection, accountId);
```
vgrichina, line 60: generally should be possible to use `await ctx.near.account()` instead of this method.

test/2fa.test.js
```diff
@@ -0,0 +1,263 @@
...
+    blockNumber = String(blockNumber);
+    const message = Buffer.from(blockNumber);
+    const signedHash = await inMemorySigner.signMessage(message, accountId);
+    const blockNumberSignature = Buffer.from(signedHash.signature).toString('base64');
+    return { blockNumber, blockNumberSignature };
+}
+
+const { parseSeedPhrase } = require('near-seed-phrase');
+const SEED_PHRASE = 'shoot island position soft burden budget tooth cruel issue economy destroy above';
+const keyPair = nearAPI.KeyPair.fromString(parseSeedPhrase(SEED_PHRASE).secretKey);
```
vgrichina, line 68: any reason to not use just random key pair?

test/2fa.test.js
```diff
@@ -0,0 +1,263 @@
...
+
+const keyStore = new nearAPI.keyStores.InMemoryKeyStore();
+
+const twoFactorMethods = [
+    { kind: '2fa-email', detail: 'hello@example.com', publicKey: 'pkemail' },
+    { kind: '2fa-phone', detail: '+1 717 555 0101', publicKey: 'pkphone' },
+];
+
+const inMemorySigner = new nearAPI.InMemorySigner(keyStore);
+async function signatureFor(accountId, valid = true) {
```
vgrichina, line 56: makes sense to extract `signatureFor` and reuse between tests.

test/2fa.test.js
```diff
@@ -0,0 +1,263 @@
...
+
+    test('generate a deterministic public key for an account', async () => {
+        await createNearAccount(accountId);
+
+        await request.post('/2fa/getAccessKey')
+            .send({
+                accountId,
+                ...(await signatureFor(accountId))
+            })
+            .expect('Content-Type', /json/)
```
vgrichina, line 98: are these `.expect` checks playing well with `await`? Wonder how errors get propagated.

test/2fa.test.js
```diff
@@ -0,0 +1,263 @@
...
+            .expect((res) => {
+                assert.equal(res.body.success, true);
+                assert.equal(res.body.securityCode.length, 6);
+                securityCode = res.body.securityCode;
+            })
+            .expect(200);
+    });
+
+    test('should be able to call init again with different method', async () => {
```
vgrichina, line 124: individual tests shouldn't depend on previous test success generally, otherwise it's hard to do one of them in isolation, execute tests in parallel, etc

test/2fa.test.js
```diff
@@ -0,0 +1,263 @@
...
+
+        await request.post('/2fa/verify')
+            .send({
+                accountId,
+                requestId,
+                securityCode,
+                testing: true,
+                ...(await signatureFor(accountId))
+            })
+            // .expect('Content-Type', /json/)
```
vgrichina, line 256: hmm?

test/app.testx.js
```diff
@@ -2,7 +2,6 @@ const dotenv = require('dotenv');
 dotenv.config('../.env');
```
vgrichina, line 2: hmm, why rename to `.testx.js`? What does it mean?

test/2fa.test.js
vgrichina: realized other tests were already doing it, so I guess just nevermind

test/app.testx.js
```diff
@@ -63,6 +81,119 @@ async function createNearAccount() {
...
+            .send({
+                accountId,
+                method,
+                securityCode: savedSecurityCode,
+            });
+        assert.equal(response.status, 200);
+    });
+
+    /********************************
+    Two people send recovery methods for the same account before it's created
```
vgrichina, line 123: why not `describe` instead of comment?

app.js
mattlockyer: done in upcoming commit

app.js
mattlockyer: was testing, uncommented in upcoming commit

app.js
mattlockyer: removed in upcoming commit

app.js
mattlockyer: clarified in upcoming commit

app.js
mattlockyer: using isNew flag in body, we will clear all recovery methods and set up this one. Changed "official" to "first".

app.js
mattlockyer: IMHO it was easier to destroy all recovery methods and create the one we will send link to vs. try to select all recovery methods that are NOT the one we send recovery link to.

middleware/2fa.js
mattlockyer: agree

middleware/2fa.js
mattlockyer: roger

middleware/2fa.js
mattlockyer: is there a reason for using keyStore for this?

middleware/2fa.js
mattlockyer: nice catch!

mattlockyer: agree it's so much cleaner now (commit coming)

middleware/2fa.js
mattlockyer: agree, this was a crutch, fixed in coming commit

middleware/2fa.js
mattlockyer: TBD

middleware/2fa.js
mattlockyer: Testing now and throwing 401 with message

middleware/2fa.js
mattlockyer: agree

middleware/2fa.js
mattlockyer: let's discuss how to instantiate later

middleware/2fa.js
mattlockyer: added MULTISIG_CONTRACT_HASHES

changed to
```
return MULTISIG_CONTRACT_HASHES.includes(state.code_hash);
```

middleware/2fa.js
mattlockyer: agree cleaned up

middleware/2fa.js
mattlockyer: Habit. I'll check with @Patrick1904 if this is baked into frontend and cleanup if easy

middleware/2fa.js
mattlockyer: I think leftover logic, removed.

middleware/2fa.js
mattlockyer: comments removed

middleware/2fa.js
mattlockyer: we can provide `2fa/update` endpoint, but out of scope for phase 0? @kcole16

middleware/2fa.js
mattlockyer: ;) great catch... I updated this so many times, need to zoom out a bit next time!

much cleaner now

middleware/2fa.js
mattlockyer: method is the user provided method and twoFactorMethod is the model record

middleware/2fa.js
mattlockyer: agree and I'd love to learn the cryptic Jest ctx.logs fishing for security codes done for recovery methods. Please help later!

mattlockyer: ok figured it out :)

middleware/2fa.js
mattlockyer: agree updated

middleware/2fa.js
mattlockyer: testing in JS now

mattlockyer: agreed ugly habit/pattern

middleware/2fa.js
mattlockyer: fixed

middleware/2fa.js
mattlockyer: validating user provided code now

test/app.testx.js
mattlockyer: sorry was skipping app tests when building 2fa.test.js, will fix in upcoming commits

middleware/2fa.js
mattlockyer: TBD if we need to test with env vars need to figure out how to reset modules

test/app.testx.js
mattlockyer: fixed

mattlockyer:
@vgrichina I've had to run tests in parallel using `jest test --runInBand`, see updated package.json. This may be why PR test is failing. Any objection to this?

app.js
vgrichina: hmm, seems still commented?

.gitignore
vgrichina: @mattlockyer ?

app.js
vgrichina: won't it be just `if (rm.public != publicKey) {` in a loop?

middleware/2fa.js
vgrichina: just seems like a logical place for this, `keyStore` is a place where you obtain a key. You won't need separate `keyStore` for every account in that case as well.

middleware/2fa.js
vgrichina: doesn't;t look like fixed

vgrichina:
@mattlockyer did you forget to push some of the commits? Seems like some stuff you marked as fixed is not fixed.

vgrichina:
>  I've had to run tests in parallel using jest test --runInBand

@mattlockyer hmm, did you mean to not run tests in parallel? `--runInBand` runs tests sequentially.

app.js
mattlockyer: sorry I was seeing the console errors in testing and it was confusing me as to which tests were not passing so for convenience commented out again while finishing the review edits.

uncommented again, tests passing

app.js
mattlockyer: you're right! cleaner now!

middleware/2fa.js
mattlockyer: sounds good, since each accountId requesting could be different I created a method called getKeyStore that returns a keyStore object

middleware/2fa.js
mattlockyer: TBD, have to check with @Patrick1904 make sure this isn't breaking on the frontend

test/2fa.test.js
mattlockyer: If the function is going to throw 401 'error message' for example, that comes through as 'text/plain' which fails here

test/2fa.test.js
mattlockyer: I'll see what I can do to evolve the 2fa testing and make it more isolated

mattlockyer:
> @mattlockyer hmm, did you mean to not run tests in parallel? `--runInBand` runs tests sequentially.

Yes you are right, "NOT" run in parallel. I think there's some issues around the mocking of console.log into ctx.logs or accounts that causes errors to fail across the 2 files. I've tried to debug this but found the easiest thing to do is just run them in sequence.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "start": "env $(sed 's/ # .*//' .env) supervisor app",
-    "test": "npm run lint && jest test",
+    "test": "npm run fix && jest test --runInBand",
```
vgrichina, line 8: we shouldn't do `fix` here, as it will mean it won't fail on CI when eslint comments are auto-fixable

test/2fa.test.js
```diff
@@ -0,0 +1,250 @@
...
+            .expect((res) => {
+                assert.equal(res.body.success, true);
+            })
+            .expect(200);
+    });
+
+});
+
+/********************************
+TBD if we need to test with env vars need to figure out how to reset modules
```
vgrichina, line 213: @mattlockyer not sure what do you mean. Why is this commented out? What needs to be reset?

vgrichina:
Looks like there are few open items still. Looks good to go overall though.

vgrichina:
Screen Shot 2020-08-05 at 1 58 29 AM

@kcole16 @corwinharrell  I think it makes sense to do some changes:
- don't include security code in subject, so that user has to read contents of transaction first
- subject can be more informative, smth like "Confirm transaction from sender.testnet to receiver.testnet"
- there is currently no info (even inside) on what account is sending a transaction. So it's possible to trick user into sending transaction from different account which often can be catastrophic.

package.json
mattlockyer: ahhh I didn't realize that the intent was to fail on CI, will update

test/2fa.test.js
mattlockyer: Fixed. Having trouble overriding Date.now, but it was trivial.

.gitignore
mattlockyer: corrected now

middleware/2fa.js
mattlockyer: ok

middleware/near.js
mattlockyer: not sure if that applies here because I need to check account for user provided accountId

test/2fa.test.js
mattlockyer: was using what was in other examples

PR #202, opened 2020-08-11
attach gas to confirm

mattlockyer:
This PR attaches 100 TGas to the confirm transaction in contract helper.


mattlockyer:
@vgrichina similar to: https://github.com/near/near-wallet/pull/797

I was consistently receiving "gas exceeded" error for contract-helper call to "confirm" method. This fixes it.

Tested on Render PR instance :+1:

vgrichina:
@mattlockyer I think we need to attach dynamic amount as discussed on Slack

PR #203, opened 2020-08-13
Wallet MainNet Release 2020-08-12

vgrichina:


vgrichina:
@mattlockyer does this look alright?

middleware/2fa.js
```diff
@@ -0,0 +1,303 @@
...
+    const keyStore = getKeyStore(accountId);
+    const near = await nearAPI.connect({
+        deps: { keyStore },
+        nodeUrl: process.env.NODE_URL
+    });
+    const contractAccount = new nearAPI.Account(near.connection, accountId);
+    const contract = new nearAPI.Contract(contractAccount, accountId, {
+        viewMethods,
+        changeMethods,
+    }, '100000000000000');
```
vgrichina, line 41: @mattlockyer nit: AFAIK `, '100000000000000'` is just ignored, there is no additional argument here:

https://github.com/near/near-api-js/blob/40d7f79c19174805e9c7500159f5166819305f5f/src/contract.ts#L22

middleware/2fa.js
```diff
@@ -0,0 +1,303 @@
...
+};
+
+// confirms a multisig request
+const confirmRequest = async (accountId, request_id) => {
+    const contract = await getContract(accountId);
+    try {
+        const res = await contract.confirm({ request_id });
+        return { success: true, res };
+    } catch (e) {
+        return { success: false, error: JSON.stringify(e) };
```
vgrichina, line 52: `JSON.stringify(e)` won't give good representation of error:

```
> JSON.stringify(new Error("Error message"))
< "{}"
```

middleware/2fa.js
```diff
@@ -0,0 +1,303 @@
...
+        securityCode,
+        kind: {
+            [Op.startsWith]: '2fa-'
+        },
+    }});
+    if (!twoFactorMethod) {
+        console.warn(`${accountId} has no 2fa method for the provided security code`);
+        ctx.throw(401, '2fa code not valid for request id');
+    }
+    // cannot test for requestId equality with negative integer???
```
vgrichina, line 273: @mattlockyer hmm? Maybe you are getting string instead of integer?

middleware/2fa.js
```diff
@@ -0,0 +1,303 @@
...
+    // checking requestId here with weak equality (no type match)
+    if (twoFactorMethod.requestId != requestId) {
+        console.warn(`2fa code not valid for request id: ${requestId} and account: ${accountId}`);
+        ctx.throw(401, '2fa code not valid for request id');
+    }
+    // only verify codes that are 5 minutes old (if testing make this impossible)
+    // console.log(twoFactorMethod.updatedAt, Date.now() - CODE_EXPIRY, twoFactorMethod);
```
vgrichina, line 280: nit: debug leftover?

middleware/near.js
```diff
@@ -44,6 +55,20 @@ async function checkAccountOwnership(ctx, next) {
     return await next();
 }
 
+async function checkAccountDoesNotExist(ctx, next) {
+    const { accountId } = ctx.request.body;
+    let remoteAccount = null;
+    try {
+        remoteAccount = (await ctx.near.account(accountId)).state();
```
vgrichina, line 62: This should be `await (await ctx.near.account(accountId)).state()`  (as `.state()` returns promise). I guess you just end up catching error from `.account()` call itself. (as it awaits on `.state()` https://github.com/near/near-api-js/blob/40d7f79c19174805e9c7500159f5166819305f5f/src/near.ts#L53).   Then it looks like everything can be reduced to:

```
try {
  await ctx.near.account(accountId);
  ctx.throw(403, 'Account ' + accountId + ' already exists.');
} catch (e) {
  return await next();
}
```

test/2fa.test.js
```diff
@@ -0,0 +1,251 @@
...
+
+jest.setTimeout(15000);
+
+beforeAll(async () => {
+    await models.sequelize.sync({ force: true });
+    ctx.near = await nearAPI.connect({
+        deps: { keyStore },
+        nodeUrl: process.env.NODE_URL
+    });
+    // ctx.accountId = 'testing' + Date.now();
```
vgrichina, line 32: debug leftover?

test/2fa.test.js
```diff
@@ -0,0 +1,251 @@
...
+});
+
+
+describe('code older than 5min should fail', () => {
+    let accountId = 'testing' + Date.now();
+    let method = twoFactorMethods[0];
+    let securityCode = '';
+    let requestId = -1;
+
+    test('initCode for an account, sets up 2fa method', async () => {
```
vgrichina, line 220: should also test a case where security code is not sent at all (as it previously was allowing it –  easy to introduce regression)

test/app.test.js
```diff
@@ -49,10 +46,31 @@ afterEach(() => {
...
 const keyStore = new nearAPI.keyStores.InMemoryKeyStore();
+
+const recoveryMethods = [
+    { kind: 'email', detail: 'hello@example.com', publicKey: 'pkemail' },
+    { kind: 'phone', detail: '+1 717 555 0101', publicKey: 'pkphone' },
+    { kind: 'phrase', publicKey: 'pkphrase' },
+];
+
 const inMemorySigner = new nearAPI.InMemorySigner(keyStore);
+async function signatureFor(accountId, valid = true) {
```
vgrichina, line 57: need to extract stuff common with 2fa.test.js into utility functions, etc

PR #207, opened 2020-08-19
Add endpoint to lookup accounts by public key

vgrichina:


middleware/indexer.js
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    return client;
+}
+
+async function findAccountsByPublicKey(ctx) {
+    const { publicKey } = ctx.params;
+
+    const client = await getPgClient();
+    const { rows } = await client.query('SELECT account_id FROM access_keys WHERE public_key = $1', [publicKey]);
```
mattlockyer, line 19: All validation/sanitization of `publicKey` is handled by `pg` library?

Not sure where this is being used but maybe we need to throw "Invalid publicKey" 400 or 422?

vgrichina: there is no sanitization needed, this is prepared SQL statement (i.e. you cannot just inject some stuff)

vgrichina:
> Looks good. Is this being used from wallet and removing wallet querying indexer directly?

eventually. For now I just need to show demo to Trust Wallet and make sure it works for them. Then we can migrate to as long-term explorer API going to be deprecated.

PR #210, opened 2020-08-21
Update Email Template and 2FA Messaging

mattlockyer:
PR updates utils/email.js to include the new template and creates 2 helper functions for recovery link emails and 2FA emails.

Messaging is according to: https://github.com/near/near-wallet/issues/751
- actions are broken out and simplified from their JSON
- args from function call are parsed from base64 and remain JSON (we won't have schema for random function calls)

Tested:
- Recovery
- Single actions
- Multiple actions (rotate keys)

Kendall has approved screenshots. Please try against your local wallet:
https://near-contract-helper-2fa.onrender.com


middleware/2fa.js
```diff
@@ -4,11 +4,12 @@ const crypto = require('crypto');
 const nacl = require('tweetnacl');
 const { creatorKeyJson } = require('./near');
 const { sendSms } = require('../utils/sms');
-const { sendMail } = require('../utils/email');
+const { sendMail, get2faHtml } = require('../utils/email');
 const models = require('../models');
 const Sequelize = require('sequelize');
 const Op = Sequelize.Op;
 const password = require('secure-random-password');
+var escape = require('escape-html');
```
vgrichina, line 12: @mattlockyer nit: why `var` here?

mattlockyer: copy from their docs, updated.

Issue #213, opened 2020-08-25
Extract HTML/text templates into separate files using templating engine (e.g. mustache)


vgrichina:
https://github.com/near/near-contract-helper/pull/210 added HTML e-mails. Makes sense to extract them from JS to make it easier to edit them, add translations, etc.

https://github.com/janl/mustache.js is pretty solid solutions for this.

PR #214, opened 2020-08-25
Fix Account Creation Emails (now with code + recovery)

mattlockyer:
This PR includes several improvements to the account creation emails.

The text versions are also updated.

This has been tested with @Patrick1904 latest wallet fix PR for account creation: https://github.com/near/near-wallet/pull/841
 
As per @corwinharrell new design and copy:
image

### DIFF
--- a/app.js
+++ b/app.js
@@ -241,7 +241,7 @@ const completeRecoveryInit = async ctx => {
 
     let publicKey
     if (seedPhrase) {
-        ({ publicKey } = parseSeedPhrase(recoverySeedPhrase))
+        ({ publicKey } = parseSeedPhrase(seedPhrase))
     }
 
     let [recoveryMethod] = await account.getRecoveryMethods({ where: {
--- a/app.js
+++ b/app.js
@@ -322,38 +322,6 @@ router.post('/account/validateSecurityCodeForTempAccount',
     completeRecoveryValidation
 );
 
-// router.post('/account/sendRecoveryMessage', async ctx => {
-//     const { accountId, method, seedPhrase, isNew } = ctx.request.body;
-//     // Verify that seed phrase is added to the account
-//     const { publicKey } = parseSeedPhrase(seedPhrase);
-//     const nearAccount = await ctx.near.account(accountId);
-//     const keys = await nearAccount.getAccessKeys();
-//     if (!keys.some(key => key.public_key === publicKey)) {
-//         ctx.throw(403, 'seed phrase doesn\'t match any access keys');
-//     }
-//     const account = await models.Account.findOne({ where: { accountId } });
-//     const [recoveryMethod] = await account.getRecoveryMethods({ where: {
-//         kind: method.kind,
-//         detail: method.detail
-//     }});
-//     await recoveryMethod.update({ publicKey, securityCode: null });
-//     if (isNew) {
-//         // clear all methods that may have been added by other users attempting to set up the same accountId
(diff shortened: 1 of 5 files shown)


mattlockyer:
@vgrichina I will definitely do #213 after this PR to clean up the template. I think this is priority for now though.

mattlockyer:
Had to pull back to draft. Realized tests will fail with changes to account flow.

app.js
```diff
@@ -271,15 +283,32 @@ const completeRecoveryValidation = async ctx => {
...
         securityCode: securityCode
     }});
 
     if (!recoveryMethod) {
-        ctx.throw(401);
+        ctx.throw(401, 'recoveryMethod does not exist');
+    }
+
+    // for new accounts, clear all other recovery methods that may have been created
+    if (ctx.request.url.indexOf('ForTempAccount') > -1) {
```
vgrichina, line 301: @mattlockyer note that the way this check is done means I can make request to `/initializeRecoveryMethod?thisPartIsIgnored=ForTempAccount` and so effectively bypass a check that account doesn't exist before.

vgrichina: There are 2 safe ways to do this:
1) set whatever option you need on `ctx.` in `checkAccountDoesNotExist`
2) have a wrapper function which would take it, so that usage of middleware would be smth like:
```
router.post('/account/validateSecurityCodeForTempAccount',
    checkAccountDoesNotExist,
     completeRecoveryValidation({ isNew: true })
 );
```

app.js
```diff
@@ -271,15 +283,32 @@ const completeRecoveryValidation = async ctx => {
...
     const account = await models.Account.findOne({ where: { accountId } });
 
+    if (!account) {
+        ctx.throw(401, 'account does not exist');
+    }
+
     const [recoveryMethod] = await account.getRecoveryMethods({ where: {
         kind: method.kind,
         detail: method.detail,
         securityCode: securityCode
```
vgrichina, line 293: @mattlockyer looks like I can just avoid supplying `securityCode` to bypass this check.

vgrichina: Looks like it doesn't work in practice because `sequelize` doesn't treat `undefined` value as absent key here. 

However I don't think it's safe to rely on this implicit behavior. We should have an explicit test for no security code supplied. I believe we discussed this before.

vgrichina:
@mattlockyer @kcole16 currently the subject for e-mail reads as:

> Your NEAR Wallet security code is: 205804

This breaks our recovery flow, which asks users to search for e-mail with "Important: Near Wallet Recovery Email" in subject. It also suggests that e-mail can be removed without reading (which it totally shouldn't).

kcole16:
@vgrichina Good point! @mattlockyer We should change this to the original recovery message title. Generally like the idea of including the verification code, but given this is a strange case we probably shouldn’t.

app.js
mattlockyer: Great catch.

Using solution 2.

app.js
mattlockyer: added: `if (!securityCode || isNaN(parseInt(securityCode, 10)) || securityCode.length !== 6)`

mattlockyer:
Updated subject and addressed other comments @kcole16

Thanks for feedback @vgrichina :+1:

app.js
vgrichina: @mattlockyer this needs test (similar as wrong code).

PR #217, opened 2020-08-28
deprecate sendRecoveryMessage

mattlockyer:
We no longer user this method from the wallet.

Part of ongoing maintenance re: account creation updates.

### DIFF
--- a/app.js
+++ b/app.js
@@ -138,64 +138,10 @@ router.post(
     }
 );
 
-router.post(
-    '/account/resendRecoveryLink',
-    checkAccountOwnership,
-    async ctx => {
-        const { accountId, seedPhrase, publicKey, method } = ctx.request.body;
-        const account = await models.Account.findOne({ where: { accountId } });
-
-        const [recoveryMethod] = await account.getRecoveryMethods({ where: {
-            kind: method.kind,
-            detail: method.detail,
-            publicKey: method.publicKey
-        }});
-
-        await recoveryMethod.update({ publicKey });
-
-        await sendRecoveryMessage({
-            accountId,
-            method,
-            seedPhrase
-        });
-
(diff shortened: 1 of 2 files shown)


middleware/2fa.js
```diff
@@ -18,7 +18,7 @@ const changeMethods = ['confirm'];
 const DETERM_KEY_SEED = process.env.DETERM_KEY_SEED || creatorKeyJson.private_key;
 const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES ? process.env.MULTISIG_CONTRACT_HASHES.split() :['7GQStUCd8bmCK43bzD8PRh7sD2uyyeMJU5h8Rj3kXXJk','AEE3vt6S3pS2s7K6HXnZc46VyMyJcjygSMsaafFh67DF', '45gayUD7tUKFc3vvGwzoPEeFS6RjdQWu7SHGpxAJe13F'];
 const CODE_EXPIRY = 300000;
-
+const oneHunTgas = '100000000000000'
```
vgrichina, line 21:
```
const NEAR_2FA_GAS = process.env. NEAR_2FA_GAS ||  '100000000000000' 
```

let's:
1) name based on purpose, not current value
2) use ALL_CAPS as customary for other constants.
3) allow to override in environment variable

Also let's not mix this into PR called "deprecate sendRecoveryMessage", as it's off-topic change.

vgrichina:
@mattlockyer check CI

middleware/2fa.js
mattlockyer: Got it, wasn't aware I was still on branch. Reverted!

mattlockyer:
@vgrichina reverted, commented, should be good!

PR #220, opened 2020-09-01
Update email subjects

mattlockyer:
This PR updates email subject lines for both regular transactions, including the receiver_id of the transaction and a warning subject line for transactions where a full access key will be added to the account.

### DIFF
--- a/middleware/2fa.js
+++ b/middleware/2fa.js
@@ -89,7 +89,7 @@ const sendCode = async (ctx, method, twoFactorMethod, requestId = -1, accountId 
         });
         
     }
-    let subject = `NEAR Wallet security code: ${securityCode}`;
+    let subject = `Confirm Transaction with ${ receiver_id }`;
     let text = `
 NEAR Wallet security code: ${securityCode}\n\n
 Important: By entering this code, you are authorizing the following transaction:\n\n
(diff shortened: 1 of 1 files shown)


middleware/2fa.js
```diff
@@ -87,9 +87,8 @@ const sendCode = async (ctx, method, twoFactorMethod, requestId = -1, accountId
             }
             requestDetails += '<br/>';
         });
-        
     }
-    let subject = `Confirm Transaction with ${ receiver_id }`;
+    let subject = `Confirm Transaction from: ${ accountId }${ request ? ` to: ${ request.receiver_id }` : ``}`;
```
vgrichina: @mattlockyer when do we have `request == null`? Shouldn't it always be available for 2FA?

mattlockyer: If the account is being set up and we are confirming the 2fa method with a code (there is no request) then requestId is -1 and request will be null.

vgrichina: > If the account is being set up and we are confirming the 2fa method with a code (there is no request) then requestId is -1 and request will be null.
 
@mattlockyer hmm, what transaction we are confirming in that case? Deploying 2FA contract?

mattlockyer:
Yes setting up account.

vgrichina:
> Yes setting up account.

So I think it makes sense to have different subject then? Something like "Confirm enabling 2FA for account_name"? cc @kcole16

vgrichina:
@mattlockyer going to merge for now, let's improve in follow-up

PR #222, opened 2020-09-02
Add a kludge to index accounts by public key

vgrichina:


frol:
Why do you need it? The Indexer does it and it already dumps the keys to the `near_indexer_for_wallet_mainnet` DB on Render. What is the point?

mattlockyer:
> Why do you need it?

It appears the indexer is not getting keys added / deleted by other contracts e.g. account created and keys added from a linkdrop.

See #indexer (slack) for more context.

frol:
We cleared up the confusion offline. Explorer database is kept synced the old way, so you won't see anything new if you query Explorer WAMP RPC. Indexer for Wallet database is Postgres instance running on Render.com is configured specifically for Wallet needs has all the necessary keys and is up and running for on mainnet for about 2 weeks now with no downtimes and lags. @khorolets is setting up the Indexer for Wallet for testnet now (it may take a few days to sync up with the network...)

You can add the endpoint reading from the database to the helper or build another API server (whichever you find a better option, it is up to the Wallet team).

Indexer for Wallet records the NEAR native AddKey and RemoveKey actions from the receipts and does it reliably for all the finalized blocks in real-time.

vgrichina:
> You can add the endpoint reading from the database to the helper or build another API server (whichever you find a better option, it is up to the Wallet team).

we have this endpoint. We just didn't have indexer working for testnet specifically apparently. Will have this change gated by env var to make sure we can use either source of data depending on whether indexer is alive for specific network.

PR #244, opened 2020-09-29
Fix contract args amount, deposit

mattlockyer:
This PR formats nested amounts and deposits in contract args.


middleware/2fa.js
```diff
@@ -78,8 +78,14 @@ const sendCode = async (ctx, method, twoFactorMethod, requestId = -1, accountId
         const { receiver_id, actions } = request;
         requestDetails = [];
         actions.forEach((a) => {
+            // format amount, deposit args
+            if (a.args) {
+                a.args = JSON.parse(Buffer.from(a.args, 'base64').toString());
```
vgrichina, line 83: @mattlockyer there is no guarantee that this is JSON.

I think it'll be best if you rebase on top of https://github.com/near/near-contract-helper/pull/237/files – it already has hexdump for binary args

vgrichina: needs tests as well (that's why https://github.com/near/near-contract-helper/pull/237/files not submitted yet)

mattlockyer:
Adding changes to #237

PR #288, opened 2020-12-09
New Indexer Updates: accounts & staking txs

mattlockyer:
Updates to indexer.js
- new indexer (ask Frol/Bohdan to share creds in lastpast)
- connects accounts endpoint to new indexer
- removes kludge accounts fix cause indexer is working now
- adds staking txs endpoint - gets all staking txs for accountId


app.js
```diff
@@ -72,10 +72,11 @@ router.post('/account', async ctx => {
 });
 
 
-const { findAccountsByPublicKey, findAccountsByPublicKeyIndexer } = require('./middleware/indexer');
-// TODO: Remove kludge when indexer returns up to date data
+const { findAccountsByPublicKey, getStakingTxs } = require('./middleware/indexer');
```
vgrichina, line 75: nit: let's stick to either `find` or `get` for consistency

middleware/indexer.js
```diff
@@ -13,24 +12,37 @@ async function getPgClient() {
...
-async function findAccountsByPublicKeyIndexer(ctx) {
-    const { publicKey } = ctx.params;
+async function getStakingTxs(ctx) {
+    const { accountId } = ctx.params;
 
     const client = await getPgClient();
-    const { rows } = await client.query('SELECT DISTINCT account_id FROM access_keys WHERE public_key = $1', [publicKey]);
-    ctx.body = rows.map(({ account_id }) => account_id);
+    const { rows } = await client.query(`
+        select
```
vgrichina, line 20: nit: I think it makes sense to format query a bit for easier reading, something like:

```
    select
         receiver_account_id validator_id,
         args->>'method_name' method_name,
         coalesce(nullif(args->>'deposit', '0'), convert_from(decode(args->>'args_base64', 'base64'), 'utf8')::jsonb->>'amount', '0') amount,
         included_in_block_timestamp ts
    from receipts
    inner join action_receipt_actions using(receipt_id)
    where
         predecessor_account_id = $1 and
         action_kind = 'FUNCTION_CALL' and 
         args->>'method_name' in ('stake', 'deposit_and_stake', 'unstake', 'unstake_all')
    limit 9999
```

middleware/indexer.js
```diff
@@ -13,24 +12,37 @@ async function getPgClient() {
...
+        args->>'method_name' method_name,
+        coalesce(nullif(args->>'deposit', '0'), convert_from(decode(args->>'args_base64', 'base64'), 'utf8')::jsonb->>'amount', '0') amount,
+        included_in_block_timestamp ts
+        from receipts
+        inner join action_receipt_actions using(receipt_id)
+        where
+        predecessor_account_id = $1 and
+        action_kind = 'FUNCTION_CALL' and 
+        args->>'method_name' in ('stake', 'deposit_and_stake', 'unstake', 'unstake_all')
+        limit 9999
```
vgrichina, line 31: 10k rows might be actually pretty expensive to query in one go, is there any reason limit has to be so high? vs say 500 which would be 10x current validators

middleware/indexer.js
```diff
@@ -13,24 +12,37 @@ async function getPgClient() {
...
 
     const client = await getPgClient();
-    const { rows } = await client.query('SELECT DISTINCT account_id FROM access_keys WHERE public_key = $1', [publicKey]);
-    ctx.body = rows.map(({ account_id }) => account_id);
+    const { rows } = await client.query(`
+        select 
+        receiver_account_id validator_id,
+        args->>'method_name' method_name,
+        coalesce(nullif(args->>'deposit', '0'), convert_from(decode(args->>'args_base64', 'base64'), 'utf8')::jsonb->>'amount', '0') amount,
+        included_in_block_timestamp ts
```
vgrichina, line 24: should the query also order in descending order by timestamp so that when it hits `limit` – we pick the most recent records?

middleware/indexer.js
mattlockyer: debatable that there will even be that many per account, it's just an upper bound so it's not too expensive.

middleware/indexer.js
mattlockyer: Consider someone stakes with one validator, does 9999 txs with other validators, then comes back and unstakes a portion from the original validator.

The wallet would have no clue how to calc rewards.

vgrichina: We are making this query over network and handle result in JS. It's not going to work fast with 10k records. Postgres going to be much faster sorting and throwing away unneeded data than client-side.

> Consider someone stakes with one validator, does 9999 txs with other validators, then comes back and unstakes a portion from the original validator.

Sounds like we need to group by validator then? What do we need to get logically – last transaction with every validator?

middleware/indexer.js
vgrichina: 10k is generally too expensive though

middleware/indexer.js
mattlockyer: Each staking tx is only a piece of the overall deposit puzzle. Ideally, we need them all to be sure that we're not missing some amount user deposited with validator long ago. We could feed in list of all validators to narrow scope of txs down.

mattlockyer:
@vgrichina do you want me to deprecate `findStakingTxs` as we no longer use it in the wallet?

Or should we keep it for now in case requirements change on wallet side?

vgrichina:
> @vgrichina do you want me to deprecate `findStakingTxs` as we no longer use it in the wallet?
> 
> Or should we keep it for now in case requirements change on wallet side?

I've removed it after merging with `master`, we can always bring it back from history if needed later.

PR #297, opened 2020-12-23
add sep env var and client for wallet indexer

mattlockyer:


middleware/indexer.js
```diff
@@ -1,10 +1,10 @@
 const { Client } = require('pg');
 
 let client;
-async function getPgClient() {
-    if (!client) {
+async function getPgClient(useWalletIndexer = false) {
+    if (!client || useWalletIndexer) {
```
vgrichina, line 5: @mattlockyer let's just have 2 separate methods and 2 separate cached clients. Like this already has a bug: after passing `useWalletIndexer = true` once – `getPgClient` would always return wallet indexer

PR #303, opened 2021-01-06
testing account activity endpoint

mattlockyer:


app.js
```diff
@@ -85,10 +85,11 @@ router.post('/account', ratelimit({
 });
 
 
-const { findAccountsByPublicKey, findStakingDeposits } = require('./middleware/indexer');
+const { findAccountsByPublicKey, findStakingDeposits, findAccountActivity } = require('./middleware/indexer');
 
 router.get('/publicKey/:publicKey/accounts', findAccountsByPublicKey);
 router.get('/staking-deposits/:accountId', findStakingDeposits);
+router.get('/account-activity/:accountId/:offset?/:limit?', findAccountActivity);
```
vgrichina, line 92: nit: I think from REST POV kinda makes sense to structure endpoints more like:

```
/account/:accountId/staking-deposits
/account/:accountId/activity
```

stuff like `offset` / `limit` should better go into query parameters

also given that we likely would want to get latest transactions first and then maybe load older transactions (and new might be appearing in between) – it's best to not rely on `offset`. Best to have parameters to specify block timestamp range, so that it's possible to implement pagination resistant to new data coming in.

middleware/indexer.js
```diff
@@ -44,6 +44,37 @@ async function findStakingDeposits(ctx) {
     ctx.body = rows;
 }
 
+async function findAccountActivity(ctx) {
+    const { accountId } = ctx.params;
+    let { offset, limit = 10 } = ctx.request.query;
+    if (!offset) {
```
vgrichina, line 50: nit: let's not call this offset, as it's more like `startingFromTimestamp`.

PR #340, opened 2021-03-07
add endpoint to return all staking pools

mattlockyer:
The current hack in the wallet of using 
https://github.com/near/near-wallet/blob/90476414962b2723ace331d19abe616dfa809b86/src/utils/staking.js#L295-L298
Does not work for validators that have **never** been included in the validating set e.g. new validators like "binancestaking.poolv1.near"

This endpoint that finds all accounts matching the "poolv1" subaccount naming, since they deployed a staking pool factory.

This should be used by the wallet and replace the above hack.

Endpoint:
```
router.get('/stakingPools/all', findAllStakingPools);
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -85,8 +85,9 @@ router.post('/account', ratelimit({
 });
 
 
-const { findAccountsByPublicKey, findStakingDeposits, findAccountActivity } = require('./middleware/indexer');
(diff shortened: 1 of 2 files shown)


vgrichina (approved):
@mattlockyer LGTM, but needs to be rebased on latest `master`

MaximusHaximus:
Closing this in favor of #478 which also works with testnet pools :+1:

PR #362, opened 2021-04-02
Rainbow bridge / Twilio message size limit / Test suites clean-up

MaximusHaximus:
Fixes https://github.com/near/near-wallet/issues/1563

Resolves issue of SMS message content being too large for twilio's 1600 character limit when sending an SMS about a rainbow bridge transaction (which contain a large hex payload).  Any `args` that are being spooled to the message content will be truncated @ 250 characters length from now on.  The new truncation logic applies only when the text is being composed for SMS delivery (emails will still get the full blob).

I identified several assertions that were throwing errors but those errors were not being caught by Jest due to sandboxing behavior combined with `instanceof` checks inside of `supertest` and other packages.  Since we had manually set Jest to run tests in series anyway, and this isn't a frontend, I replaced Jest with Mocha/Chai/Sinon, which solves the sandboxing issue and gives us a more standard toolset for a backend/node project.

Added an estlintrc file just for the `test` directory so that eslint doesn't think we have globals in scope in non test files

Removed the console.log->array of strings->regex parsing pipeline to get security codes into the test suites scope for use.  Since the `app` object is an EventEmitter, and it's on `ctx` (koa context), I just updated the code that used to `console.log` to emit server events whenever we create a security code.  Also added events for `verbose output mode` that can easily allow the integration test suites to print out the details of email and sms messages generated by the test suite (also leveraging event emitter emissions on the koa server)

Added a `constants.js` and moved a bunch of magic strings into that file so it's clear when they are referenced that they are logically the same thing across the codebase, and what that thing is, without reading the surrounding context.

Created a 2fa service class that contains the methods from the `2fa` module.  A singleton instance of that class is bound to koa route handlers in `middleware/2fa.js`.  Having the 2fa code encapsulated in this class will allow easier testing of stuff like expired codes in the future.

Heavily refactored the composition of messages that was being done in 2fa.js, so that the code reads linearly and the message content is composed in a single place rather than across a sequence of mutations of `text`, `subject`.  

The above-mentioned message content composition methods are tested with acceptance tests that spool the content of both text/subject and html to files for easy review when they change.   See the files in `test/message_formatting/samples/html` and `test/message_formatting/samples/text` to see what this looks like in practice.  Any time the output of these changes, the dev who made the change will be prompted to re-create the sample files to accept the changes.

I fixed a bug that Vlad pointed out -- content being sent in `requestContent` that was being included in the `text` for SMS messages was being HTML escaped, when it didn't need to be.  Content being rendered for use in an e-mail _will_ still be HTML escaped.

Enhanced `formatArgs()` so that `amount` and `deposit` are always the first properties printed when `args` is serialized and is JSON parseable. Since we are now truncating `args` paylods @ 250 chars for SMS delivery, putting these properties first when serialized makes it much more likely that they will be visible in the transactions content sent in the SMS message.

Replaced a lot of manual gluing of strings with template strings for readability's sake

Added a dedicated `test/.env` and removed the manual mutation of `process.env` at the top of the integration test files.  Imports into test files are no longer sensitive to the order they are loaded... ;)

Fixed pluralization of 'transaction(s)' in a few spots

Fixed account ID _never_ being included in the subject line of 2fa verification e-mails

Encapsulated the interactions between the test server (app, security key extractions), keyPair, keyStore, signer,  and NEAR API client into a class called `TestAccountHelper`.  Re-used this helper in both integration test suites to eliminate boilerplate both in set-up and in interactions with the test server.

Once I added assertions to the existing `app` integration tests that actually checked the returned message content rather than just HTTP status codes, I found that several of them are actually not testing what the test descriptions appear to think they are testing; the message returned indicates that while the HTTP code matches the assertion, the reason why that HTTP code was returned is a different one.  This PR got pretty huge already, so I marked those as FIXME to come back around and fix in a second pass; I didn't want to change the behavior of existing test suites in non-obvious ways while still working on the test framework switch to Mocha.


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "scripts": {
     "start": "env $(sed 's/ # .*//' .env) supervisor app",
-    "test": "npm run lint && jest test --runInBand",
+    "test": "yarn lint && mocha",
```
MaximusHaximus, line 8: Noticed this was still using `npm`, so updated it to use `yarn` for consistency...

MaximusHaximus:
@kcole16 @jimmy3dita See files in `samples` directory for reference of what our current message content is (both for email and SMS texts) for various flows.  The HTML output files can even be viewed in Chrome as a normal webpage.

jimmy3dita (approved):

vgrichina (changes requested):
Looks good overall, however as it is now it's pretty big an hard to review thoroughly.

Few ideas on how to simplify:
1) Separate into few different PRs (test conversion, refactor, bugfix itself)
2) Avoid doing code style changes in the same PR. 

E.g. minor changes everywhere like changing `if (method.kind === 'phone') {` to ` if (method.kind === RECOVERY_METHOD_KINDS.PHONE) {` everywhere makes it extremely hard to review substantial changes.

constants.js
```diff
@@ -0,0 +1,23 @@
+const SERVER_EVENTS = {
```
vgrichina, line 1: I'd prefer we don't make global constants file, but scope constants to the relevant modules. So that e.g. `TWO_FACTOR_AUTH_KINDS` live somewhere close to 2fa implementation.

middleware/2fa_service/TwoFactorAuthService.js
```diff
@@ -0,0 +1,361 @@
...
+    // https://github.com/near/core-contracts/blob/636e7e43f1205f4d81431fad0be39c5cb65455f1/multisig/src/lib.rs
+    '8DKTSceSbxVgh4ANXwqmRqGyPWCuZAR1fCqGPXUjD5nZ',
+    // https://github.com/near/core-contracts/blob/f93c146d87a779a2063a30d2c1567701306fcae4/multisig/res/multisig.wasm
+    '55E7imniT2uuYrECn17qJAk9fLcwQW4ftNSwmCJL5Di',
+];
+
+const CODE_EXPIRY = 30 * 60000;
+const GAS_2FA_CONFIRM = process.env.GAS_2FA_CONFIRM || '100000000000000';
+
+class TwoFactorAuthService {
```
vgrichina, line 41: Why create class to represent stateless middleware and helper functions?

vgrichina: It's best to revert this change, so that changes to `2fa.js` are reviewable.

middleware/2fa_service/index.js
```diff
@@ -0,0 +1,4 @@
+const TwoFactorAuthService = require('./TwoFactorAuthService');
+
+// TODO: Provide dependencies / environment variables as constructor arguments?
+module.exports = new TwoFactorAuthService();
```
vgrichina, line 4: not sure why instead of 2fa.js we now have 3 modules to achieve exactly the same thing (2 of these modules being few lines of code like this one).

middleware/2fa_service/messageContent.js
```diff
@@ -0,0 +1,139 @@
...
+        return escapeHtmlIfNotSMS(`Staking: ${fmtNear(amount)} to validator: ${receiver_id}`);
+    case 'AddKey':
+        if (permission) {
+            const { allowance, receiver_id, method_names } = permission;
+            const methodsMessage = method_names && method_names.length > 0 ? `${method_names.join(', ')} methods` : 'any method';
+            return escapeHtmlIfNotSMS(`Adding key ${public_key} limited to call ${methodsMessage} on ${receiver_id} and spend up to ${fmtNear(allowance)} on gas`);
+        }
+        return escapeHtmlIfNotSMS(`Adding key ${public_key} with FULL ACCESS to account: "${receiver_id}"`);
+    case 'DeleteKey':
+        return escapeHtmlIfNotSMS(`Deleting key ${public_key}`);
```
vgrichina, line 74: nit: seems like it makes sense to escape HTML once in the caller of this function instead of doing it here for every return option

middleware/2fa_service/messageContent.js
```diff
@@ -0,0 +1,139 @@
...
+
+const formatArgs = (args, isForSms) => {
+    let output = '';
+
+    const argsBuffer = Buffer.from(args, 'base64');
+    try {
+        const jsonString = argsBuffer.toString('utf-8');
+        const parsed = JSON.parse(jsonString);
+
+        // Composing new obj in specific order so `amount` and `deposit` will likely be in the first 250chars
```
vgrichina, line 17: this seems to be some new logic, what is it's purpose?

if this is trying to format NEAR amounts in method call args, we shouldn't do it as args can be anything we cannot be sure that `amount` there is actual NEAR tokens and so can misrepresent what is going on

test/toBase64.js
```diff
@@ -0,0 +1,145 @@
+// Adapted from https://github.com/beatgammit/base64-js
+// MIT license
+
+function Base64Thing() {
```
vgrichina, line 4: why do we need base64 reimplementation???

test/TestAccountHelper.js
```diff
@@ -0,0 +1,219 @@
...
+    async ensureNEARApiClientInitialized() {
+        if (!this._NEARApiClient) {
+            this._NEARApiClient = await nearAPI.connect({
+                deps: { keyStore: this._keyStore },
+                nodeUrl: process.env.NODE_URL
+            });
+        }
+    }
+
+    get near() {
```
vgrichina, line 49: not sure I understand why define explicit getters for things where you can expose property directly

test/TestAccountHelper.js
```diff
@@ -0,0 +1,219 @@
...
+const { SERVER_EVENTS } = constants;
+
+function extractValueFromHash(hash, key) {
+    const value = hash[key];
+    delete hash[key];
+
+    return value;
+}
+
+class TestAccountHelper {
```
vgrichina, line 14: extracting stuff here might be a good move, but there is way too many huge changes at the same time to review productively

MaximusHaximus:
@vgrichina thanks for the review! I'll get working on splitting this up into multiple PRs per your request :+1:

jimmy3dita:
@MaximusHaximus @vgrichina the Rainbow Bridge team tested this change on Testnet, confirming it works as expected.

PR #363, opened 2021-04-07
Test cleanup / Jest -> Mocha/Chai/Sinon

MaximusHaximus:
**1. Replaces Jest with Mocha/Chai/Sinon**
- Added @types for the new packages for nicer dev experience
- Added custom eslintrc for `test` directory so that eslint doesn't believe that mocha globals exist in our main codebase files
- Updated README.md referencing Jest accordingly
- Modified package.json `test` task entry to use `yarn` to run lint instead of `npm`

**2. Cleans up integration test files for `2fa` and `app`**
- **Consolidated a bunch of shared setup and boilerplate** into `test/TestAccountHelper` class that is responsible for the interactions between the test server, keypair, keystore, etc.
- Replaced manual mutation of `process.env` with a dedicated `test.env` dotEnv file; the existing code used `dotEnv` to load _some_ things, but then manually mutated `env` with others, making it hard to reason about what source of truth the tests were actually using.
- **Replaced the need to monkey-patch `console.log` in the test files and parse specific magic string shapes to extract `securityCode` from the console.log outputs generated during an integration test run.**  This is now being accomplished using an event emitter on the server that emits `SECURITY_CODE` events any time a new code is generated.  Now we can get security codes during test runs without having to parse message content (which required specific matching depending on the message being generated, e.g. 2fa vs. approval code are different...) and our tests won't break if we re-word our message content in the future.
- Added `expectRequestHelpers` file with some basic assertion helpers.  This is necessary as soon as you return a `supertest` chain across an `async` boundary, but still want to do some assertions on the chain in the caller scope, because the request is _sent_ when anything calls `.then()` on it, so can't have any further `.expect()` calls added to it after being returned from an `async` function
- Added `createTestServerInstance` that is a simple method create a supertest instance from the `app` and return both the app and the supertest `request` handle.  At some point it'll be useful to make this file be call a factory method with customized route bindings for test purposes (for example, right now some of our tests assume that `123123` will never actually be the securityCode that we generate internally...but because we're using `password` to create the `securityCode`, it _could_ be `123123` is actually the correct security code.  With server composition being done as a factory, we could create an instance of our `2fa` or `recoveryMethods` service with a shimmed version of the code that creates securityCode so that this test case could be deterministic, where it is not currently.

3. Fixes a bug wherein email messages sent from the `initCode` flow _never_ had `accountId` in scope, so subject lines were always incorrect.
- I stumbled across this while adding `emit()` code to capture security code generation, because I always track security codes by their associated `accountId`, but no `accountId` was available when being called from the `initCode` flow.

### DIFF
--- a/middleware/2fa.js
+++ b/middleware/2fa.js
@@ -1,14 +1,20 @@
 const nearAPI = require('near-api-js');
-const { sendSms } = require('../utils/sms');
-const { sendMail, get2faHtml } = require('../utils/email');
-const models = require('../models');
-const Sequelize = require('sequelize');
-const Op = Sequelize.Op;
+const escapeHtml = require('escape-html');
 const password = require('secure-random-password');
-const escape = require('escape-html');
+const Sequelize = require('sequelize');
+
+const constants = require('../constants');
+const models = require('../models');
+const emailHelper = require('../utils/email');
+const smsHelper = require('../utils/sms');
+
+const Op = Sequelize.Op;
+const { sendSms } = smsHelper;
+const { sendMail, get2faHtml } = emailHelper;
+const { SERVER_EVENTS, TWO_FACTOR_AUTH_KINDS } = constants;
 
 const SECURITY_CODE_DIGITS = 6;
-const twoFactorMethods = ['2fa-email', '2fa-phone'];
+const twoFactorMethods = Object.values(TWO_FACTOR_AUTH_KINDS);
 
 const MULTISIG_CONTRACT_HASHES = process.env.MULTISIG_CONTRACT_HASHES ? process.env.MULTISIG_CONTRACT_HASHES.split() : [
     // https://github.com/near/core-contracts/blob/fa3e2c6819ef790fdb1ec9eed6b4104cd13eb4b7/multisig/src/lib.rs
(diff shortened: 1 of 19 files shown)


middleware/2fa.js
```diff
@@ -219,7 +235,7 @@ const initCode = async (ctx) => {
         return;
     }
     // client waits to deploy contract until code is verified
-    await sendCode(ctx, method, twoFactorMethod);
+    await sendCode(ctx, method, twoFactorMethod, -1, accountId);
```
vgrichina, line 238: what is the reason behind this change?

test/2fa.test.js
```diff
@@ -1,257 +1,254 @@
...
+        // Would prefer beforeAll, but `testContext.logs` is cleared in the global beforeEach() that would run after beforeAll here
+        beforeEach(async () => {
+            if (!accountId) {
+                accountId = await testAccountHelper.createNEARAccount();
+            }
+        });
 
-    test('initCode for an account, sets up 2fa method', async () => {
-        await createNearAccount(accountId);
+        // TODO: Should we be cleaning up accounts just in case this is running against e.g. testnet?
```
vgrichina, line 72: we probably don't care much unless we want to run tests on mainnet

might be slightly easier for debugging failures if account is still around

probably best not to keep commented out code

test/TestAccountHelper.js
```diff
@@ -0,0 +1,219 @@
+const nearAPI = require('near-api-js');
+
+const constants = require('../constants');
+
+const { SERVER_EVENTS } = constants;
+
+function extractValueFromHash(hash, key) {
```
vgrichina, line 7: nit: given this is only used one time and essentially 2-liner, I'd just inline it.

vgrichina: P.S. Wonder why you call it hash, this brings me back to the time I was writing some hacky Perl in one of my first web apps :)

test/expectRequestHelpers.js
```diff
@@ -0,0 +1,34 @@
+const chai = require('./chai');
+
+const { expect } = chai;
+
+function expectJSONResponse(res) {
```
vgrichina, line 5: nit: wonder if these custom checks should be plugged into chai:

https://www.chaijs.com/api/plugins/

utils/email.js
```diff
@@ -15,7 +15,7 @@ const sendMail = async (options) => {
             ...options
         });
     } else {
-        console.log('sendMail:', options);
+        emitEmailSentEvent(options);
```
vgrichina, line 18: let's still make sure to log to console as it is convenient for debugging locally in `development` environment (maybe already wired elsewhere but I didn't notice?)

middleware/2fa.js
MaximusHaximus: https://github.com/near/near-contract-helper/pull/363/commits/db96886a2ce2fa61e7f35b2ec4de5497410daf5f

Prior to this, subject line for 2fa email never included account Id even though it attempted to include it.  Emit of security code events includes account Id that the security code was generated for, which highlighted the problem beyond just the subject line being broken in 2fa emails :)

vgrichina (approved):
Few minor comments, but otherwise feel free to merge

test/TestAccountHelper.js
MaximusHaximus: :+1: I guess it could have been `extractValueFromObj()` :). Dunno why my mind wrote it that way... :) In-lining.

test/expectRequestHelpers.js
MaximusHaximus: :+1: It would be nicer than the way that `supertest` works now that promises are first class citizens in JS-land.  I haven't built any custom chai plugins before, so I did it as a standalone function for now in the interest of time...

utils/email.js
MaximusHaximus: Yup, echoing this data to console is still supported; check out `test/2fa.test.js:34` and `test/app.test.js:43` -- now you can turn on echoing either security codes, or message contents, or both, from the top of the test file directly optionally if you need the extra verbosity, but during normal runs the tests don't create extra console.log output :+1:  

It's worth noting that the output is also formatted for easy readability now (no \n's visible; they are actually converted to newlines) when it _is_ console.logged during a test run; default behavior logs everything that is not HTML by default -- acceptance tests coming right after this PR give us visibility into HTML output in a much nicer way than console.logs

MaximusHaximus:
@vgrichina pushed commits to address your feedback, and responded to your question :)

PR #364, opened 2021-04-08
Add message content acceptance tests

MaximusHaximus:
1. Defines message composition helper methods outside of our REST API routes (`accountRecoveryMessageContent.js` and `middleware/2faMessageContent.js`, so that the output of our message composition logic can be verified (programmatically!) without needing to run full integration suites and comb through console.log outputs.

2. Adds a set of acceptance tests that spool message contents for each type of message, for each delivery format (HTML and not HTML), to disk in `test/messages/acceptanceTestOutputs`.
- If the code that generates the message contents here changes in such a way that the message contents changes, the tests will fail until the developer forces the acceptance test output file to be regenerated
- Regenerating the output file is accomplished by either deleting the file and re-running the suite or by setting a boolean in the test suite itself to true before re-running it.

3. Updates code in `middleware/2fa.js` and in `app.js` to use the message composition helpers.  No changes were made to  the actual content of the messages in this PR - just the code structure that generates the message content, and acceptance tests.


middleware/2faMessageContent.js
```diff
@@ -0,0 +1,99 @@
...
+        if (permission) {
+            const { allowance, receiver_id, method_names } = permission;
+            const methodsMessage = method_names && method_names.length > 0 ? `${method_names.join(', ')} methods` : 'any method';
+            return escapeHtml(`Adding key ${ public_key } limited to call ${methodsMessage} on ${receiver_id} and spend up to ${fmtNear(allowance)} on gas`);
+        }
+        return escapeHtml(`Adding key ${ public_key } with FULL ACCESS to account`);
+    case 'DeleteKey':
+        return escapeHtml(`Deleting key ${ public_key }`);
+    }
+};
```
MaximusHaximus, line 39: The code in these methods was copied _as-is_ from `middleware/2fa.js`; no changes were made in this PR.

PR #365, opened 2021-04-08
Rainbow bridge / Twilio message size limit & message content fixes

MaximusHaximus:
Fixes https://github.com/near/near-wallet/issues/1563

1. Modifies `formatArgs()` to always serialize `amount` and `deposit` as the first properties in `args`, so that when `args` is truncated, it is not likely that `amount` or `deposit` will be truncated.
2. Modified `formatArgs()` to truncate output at max 250 chars when `args` is longer than 250 chars -- but conditionally, only if the target delivery destination is SMS, which has strict length limits.  HTML output for emails is unaffected.
3. Modified `formatAction` to no longer apply HTML escaping when the target delivery destination is SMS -- @vgrichina this should fix special chars being HTML escaped in args content that we send in the SMS message per your note
4. Modified printing of multiple transactions into the HTML 2fa page so that they are printed 1-per-line, which matches how the SMS message content is formatted.  Now you can actually read multiple transactions being rendered in the HTML page at the same time! :)
5. Added acceptance test for not applying HTML and truncating args outputs for 2fa transaction approval test
6. Added specific behavioral tests for `formatActions()` and `formatArgs()` that verify that truncation and HTML escaping happen (and don't happen) when they should be.


middleware/2faMessageContent.js
```diff
@@ -2,40 +2,71 @@ const hex = require('hexer');
...
-        if (json.deposit) json.deposit = fmtNear(json.deposit);
-        return JSON.stringify(json);
-    } catch(e) {
+        const parsed = JSON.parse(jsonString);
+
+        // Composing new obj in specific order so `amount` and `deposit` will likely be in the first 250chars
+        const { amount, deposit, ...json } = parsed;
+
+        const formattedNear = {
+            ...Object.hasOwnProperty.call(parsed, 'amount') && { amount: fmtNear(amount) },
```
vgrichina, line 21: @MaximusHaximus not gonna block on this as logic is the same as before, looks like added here:
https://github.com/near/near-contract-helper/pull/244/files

Looks like it's not correct logic however, as there is no guarantee that `deposit` and `amount` fields contain NEAR tokens. Every fungible token has `amount` field and they can have arbitrary number of decimals:
https://explorer.mainnet.near.org/transactions/8MSMkTkHKy77426bTghQX9kRvwb9Xp9ssa3QzCcWggvJ

PR #367, opened 2021-04-09
Hotfix: Relax requirements for accounts matching public key

vgrichina:


middleware/indexer.js
```diff
@@ -93,7 +93,6 @@ const findAccountsByPublicKey = withPgClient(async (ctx) => {
         WHERE public_key = $1
             AND accounts.deleted_by_receipt_id IS NULL
```
frol, line 94: You don't need to join and check if an account is deleted. Indexer automatically marks the keys removed when an account gets deleted (fixed a while ago: https://github.com/near/near-indexer-for-explorer/pull/63)

frol (approved):
LGMT

PR #383, opened 2021-04-27
Funded Accounts (reCaptcha protected)

MaximusHaximus:
Backend for https://github.com/near/near-wallet/discussions/1659

**NEW ROUTE**
`POST /fundedAccount`, which will create an account, funded with some tokens from a source account.

**NEW ENVIRONMENT VARIABLES**
RECAPTCHA_SECRET - secret required to be sent to google API endpoint for each request we make
MAX_GAS_FOR_ACCOUNT_CREATE - Maximum gas that _could_ be required to complete an account create (passed directly to functionCall() method)
FUNDED_ACCOUNT_BALANCE - Initial balance for funded accounts
FUNDED_NEW_ACCOUNT_CONTRACT_NAME - target contract name where we should call `create_account` method to create a new, funded account.  Default is `near` (primary mainnet use-case)

**PACKAGE UPDATES**
Added `nyc` to dev dependencies for plug-n-play test coverage reporting (added .nyc_output to .gitignore)
Added explicit `superagent` dependency (it was already a dev dependency of `supertest`) to perform the google API request
Added `nock` dev dependency so I could write a few basic tests around the reCaptcha code without them being integration tests with www.google.com ;)


app.js
```diff
@@ -87,7 +92,7 @@ router.post('/account', ratelimit({
 }), async ctx => {
     if (!creatorKeyJson) {
         console.warn('ACCOUNT_CREATOR_KEY is not set up, cannot create accounts.');
-        ctx.throw(404);
+        ctx.throw(500, 'Service misconfigured; account creation is not available.');
```
MaximusHaximus, line 95: @vgrichina This felt more like a `500` than ` 404` since expected functionality is broken without this environment variable but to the REST client a `404` seems confusing in this context, since the request is a POST that creates a new subject... happy to revert if this is an incorrect interpretation here

RecaptchaValidator/RecaptchaValidator.js
```diff
@@ -0,0 +1,88 @@
...
+    constructor({
+        request = superagent,
+        RECAPTCHA_SECRET
+    }) {
+        this.request = request;
+        this.RECAPTCHA_SECRET = RECAPTCHA_SECRET;
+    }
+
+    // Documentation: https://developers.google.com/recaptcha/docs/verify
+    async validateRecaptchaCode(recaptchaCode, remoteIp) {
```
vgrichina, line 57: nit: consider using a function instead of immutable object with one method

```
module.exports = (request, recaptchaSecret) => (recaptchaCode, remoteIp) => {
    // basically same code here
}
```

RecaptchaValidator/RecaptchaValidator.js
```diff
@@ -0,0 +1,88 @@
+const superagent = require('superagent');
```
vgrichina, line 1: nit: note that we depend on `node-fetch` already in `near-api-js`, but also should be ok to use `superagent` if it's more convenient for the job (given it's not public js bundle)

app.js
```diff
@@ -4,16 +4,20 @@ const app = new Koa();
...
 const constants = require('./constants');
+const recaptchaValidator = require('./RecaptchaValidator');
 
 const { RECOVERY_METHOD_KINDS, SERVER_EVENTS } = constants;
 
 app.use(require('koa-logger')());
 app.use(body({ limit: '500kb', fallback: true }));
 app.use(cors({ credentials: true }));
 
-let reportException = () => {};
```
vgrichina, line 16: nit: let's try to not reformat otherwise untouched code (slightly increases review burden).

app.js
```diff
@@ -100,6 +105,66 @@ router.post('/account', ratelimit({
     ctx.body = await masterAccount.createAccount(newAccountId, newAccountPublicKey, NEW_ACCOUNT_AMOUNT);
 });
 
+// TODO: Adjust gas to correct amounts
+const MAX_GAS_FOR_ACCOUNT_CREATE = process.env.MAX_GAS_FOR_ACCOUNT_CREATE || '100000000000000';
```
vgrichina, line 109: nit: seems like good occasion to move account creation into separate module in `middleware/`

app.js
```diff
@@ -100,6 +105,66 @@ router.post('/account', ratelimit({
...
+
+    if(!newAccountPublicKey) {
+        ctx.throw(400, 'newAccountPublicKey is required');
+    }
+
+    if(!recaptchaCode) {
+        ctx.throw(400, 'recaptchaCode is required');
+    }
+
+    if (newAccountId.includes('dzarezenko')) {
```
vgrichina, line 142: no need to keep this, was a temporary protection against my friend trolling us

vgrichina (approved):

app.js
MaximusHaximus: Will do! Also, FYI, you can hide purely whitespace changes from the diff entirely in GH interface
image

RecaptchaValidator/RecaptchaValidator.js
MaximusHaximus: It would take a fair amount of code to re-create superagent's error handling and retry logic, so if this isn't a blocker I'll leave it using `superagent` here.

Also, for my understanding, I'm curious why you wouldn't consider using `superagent` for a public js bundle; is the concern around bundle size, or are there other reasons that would restrict us from using it there?

app.js
MaximusHaximus: Done!

RecaptchaValidator/RecaptchaValidator.js
MaximusHaximus: I'll give this a try next time through -- generally, I try to avoid closure scope complexity as a default; I have found that as codebases get more complex, using closures to manage scope becomes harder to reason about, easier to introduce memory leaks, harder to debug, more complex to document (either with JSDoc or when adding types), and more difficult to refactor (especially using automated tools) than using statically defined objects.  Also, IDEs have a hard time with inferring about things that are passed in dynamically using thunks; I generally see much better code completion and intellisense when things are statically defined even if it's a little bit more verbose to define them that way

PR #398, opened 2021-05-08
1659 - Funded Accounts - Update error response shape and use form-data posts for API calls

MaximusHaximus:
1. When we encounter a known error, we now return the body payload as JSON, with shape { success, result?, message?, code? }.  
- When `success` is `true`, `result` will contain the transaction result from the account creation, but `message` and `code` will not exist
- When `success` is `false`, `message` and `code` will contain details about what caused the failure
- If an unknown error is encountered while trying to create the account, it will be thrown via `ctx.throw` which returns just the stringified error in the body.

2. Updated reCaptcha POST behavior to use multi-part form posting, as the Google API endpoint doesn't accept JSON 🤦  :)

### DIFF
--- a/RecaptchaValidator/RecaptchaValidator.js
+++ b/RecaptchaValidator/RecaptchaValidator.js
@@ -72,10 +72,12 @@ class RecaptchaValidator {
             errorCodes.push(ERROR_CODES.TRANSPORT_ERROR);
         }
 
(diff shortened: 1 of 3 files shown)


MaximusHaximus:
@vgrichina Merging this as the changes are entirely in un-called code and it's blocking testing of the reCAPTCHA/funded account in a PR/branch instance of the staging wallet

PR #407, opened 2021-05-11
Speed up DB queries using updated schema

vgrichina:


middleware/indexer.js
```diff
@@ -105,8 +105,8 @@ const findReceivers = withPgClient(async (ctx) => {
     // NOTE: Looks like not doing a join is much faster (not surprising, but doesn't allow for FUNCTION_CALL filtering)
     // So potential solution might be to maintain materialized view of likely tokens and query for all receivers instead
     const { rows } = await client.query(`
-        select distinct receiver_account_id from receipts
-        join action_receipt_actions on using (receipt_id)
+        select distinct receiver_account_id
+        from action_receipt_actions
         where predecessor_account_id = $1
             and action_kind = 'FUNCTION_CALL'
```
frol, line 111: It should be `receipt_receiver_account_id` and `receipt_predecessor_account_id`

middleware/indexer.js
```diff
@@ -119,21 +119,20 @@ const findLikelyTokens = withPgClient(async (ctx) => {
 
     // TODO: Make sure indexer for explorer DB allows for faster way to do it in prod (see also above)
     const mintedWithBridge = `
-        select distinct receiver_account_id from (
-            select convert_from(decode(args->>'args_base64', 'base64'), 'UTF8')::json->>'account_id' as account_id, receiver_account_id
-            from receipts
-            join action_receipt_actions using (receipt_id)
+        select distinct receipt_receiver_account_id as receiver_account_id from (
+            select args->'args_json'->>'account_id' as account_id, receipt_receiver_account_id
```
frol, line 123: Do you want to have `as receiver_account_id`?

middleware/indexer.js
vgrichina: thx for catching untested route!

we prob not using it in wallet rn

middleware/indexer.js
vgrichina: not sure what do you mean. it's renamed in outer `select` query.

frol: nevermind, I misread the code

middleware/indexer.js
frol: It should be either removed or fixed, I believe

vgrichina: Fixed

PR #408, opened 2021-05-12
1659 - Add checkFundedAccountAvailable endpoint

MaximusHaximus:
Endpoint checks the available balance in the source account used for creating funded accounts, and returns whether or not that account has enough balance to provide a funded account to the user.

This will allow us to hide the entire reCAPTCHA flow entirely in cases where the funded account has no more remaining balance, rather than the user still being forced to solve the reCAPTCHA, only to be directed to the implicit account creation screen after they have solved it.


MaximusHaximus:
@vgrichina just an FYI

PR #422, opened 2021-05-27
Index tokens received by user

vgrichina:


MaximusHaximus (approved):
:+1: 🚢  Just curious what kind of performance increases we see with this; is it a drastic improvement?

vgrichina:
> 👍 🚢 Just curious what kind of performance increases we see with this; is it a drastic improvement?

this query in particular introduces filtering by `receiver_id` in JSON args, which lacked index before (and now is a simple lookup in index). So it more like introduces new feature vs perf improvement.

If you mean in a broader sense (i.e. in https://github.com/near/near-contract-helper/pull/407) then it depends on query by query, but generally one of main improvement is that you can filter `action_receipts_actions` by receiver / predecessor account IDs without first doing join with `receipts`.

PR #437, opened 2021-06-07
426 -  Funded account needs deposit management

MaximusHaximus:
**This PR should be merged after https://github.com/near/near-contract-helper/pull/436.**

1. Moved funded account related methods into their own file, as there are more functions unrelated to creating a funded account than managing/ checking states relating to funded accounts now.
2. Added `fundedAccountNeedsDeposit` to Account sequelize model.
3. **Route changes:**
- Updated **POST `/fundedAccount`**
  - Allows only 1 per minute per IP, to try to alleviate some account creation spamming abuse.
  - Left the existing `/createAccount` route set to 10 per 15 minutes.
- Added **GET `/account/walletState/:accountId`** 
  - Returns wallet specific state -- not NEAR on-chain account related data.
  - Shape: `{ accountId, requiredUnlockBalance, fundedAccountNeedsDeposit }`
  - requiredUnlockBalance allows the frontend to show an accurate nag screen without needing to keep constants in lockstep between the frontend and backend.
- Added **POST `/clearFundedAccountNeedsDeposit`**
  - Sets the `fundedAccountNeedsDeposit` value on an account to `false` -- but only if there are sufficient funds in the account to do so.  
  - This endpoint is idempotent, so it will return success: true if the `fundedAccountNeedsDeposit` has already been set to `false` before.
  - Returns `success: false` if the account does not have sufficient balance to be unlocked, which is controlled by a new environment variable `UNLOCK_FUNDED_ACCOUNT_BALANCE`, defaulted to 0.2 NEAR.
  - Returns `requiredUnlockBalance` in response payload if the unlock failed, so the UI can display the correct amount needed to unlock the account from the payload result instead of 'guessing' :)
- Updated **POST `/fundedAccount`**
  - Creates an Account row that has `fundedAccountNeedsDeposit` set to `true`.
  - Added `requiredUnlockBalance` to the JSON payload returned on success so the frontend UI can render the true amount required to unlock the account on a successful create.
4. Renamed some variables for clarity.
5. **`withAccount` middleware**
- Renamed to `withSequelizeAccount` for clarity, and updated it to place the account model object on `ctx.sequelizeAccount` instead of `ctx.account`.
- Wrapped it in a factory function that allows us to use it with either GET or POST routes; for GET routes we need to check `ctx.params`, whereas existing POST route needed to look for the `accountId` value on `ctx.body.accountId`.


Patrick1904 (approved):
LGTM!

app.js
```diff
@@ -134,14 +151,26 @@ router.post('/account/recoveryMethods', checkAccountOwnership, async ctx => {
     ctx.body = await recoveryMethodsFor(account);
 });
 
-async function withAccount(ctx, next) {
-    const { accountId } = ctx.request.body;
-    ctx.account = await models.Account.findOne({ where: { accountId } });
-    if (ctx.account) {
-        await next();
-        return;
+function getWithSequelizeAccountHandler(source) {
```
vgrichina, line 154: nit: I think it's pretty common to have middleware with arguments (e.g. `ratelimit` middleware), so it makes sense to call outer function `withSequelizeAccount` instead of `getWithSequelizeAccountHandler`.

middleware/createAccount.js
```diff
@@ -0,0 +1,25 @@
...
+
+const createAccount = async (ctx) => {
+    if (!creatorKeyJson) {
+        console.warn('ACCOUNT_CREATOR_KEY is not set up, cannot create accounts.');
+        ctx.throw(500, 'Service misconfigured; account creation is not available.');
+    }
+
+    const { newAccountId, newAccountPublicKey } = ctx.request.body;
+
+    if (newAccountId.includes('dzarezenko')) {
```
vgrichina, line 13: time to remove this check, he already stopped trolling us I believe :)

middleware/fundedAccount.js
```diff
@@ -119,15 +112,51 @@ const createFundedAccount = async (ctx) => {
...
+        await ctx.sequelizeAccount.update({ fundedAccountNeedsDeposit: false });
+
+        ctx.status = 200;
+        ctx.body = { success: true };
+        return;
+    }
+
+    setJSONErrorResponse({
+        ctx,
+        statusCode: 200,
```
vgrichina, line 140: let's not use 200 code for failed requests

vgrichina (changes requested):
There seems to be missing `await` in

```
const [sequelizeAccount, fundingAccount] = Promise.all([
```

this is a blocker as it suggests the code might be untested (or only worked well by incident)

app.js
```diff
@@ -134,14 +151,26 @@ router.post('/account/recoveryMethods', checkAccountOwnership, async ctx => {
...
+    if (source !== 'body' && source !== 'params') {
+        throw new Error('invalid source for accountId provided to getWithSequelizeAccountHandler()');
     }
-    ctx.throw(404, `Could not find account with accountId: '${accountId}'`);
+
+    return async function withSequelizeAccount(ctx, next) {
+        let accountId;
```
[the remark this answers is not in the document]
MaximusHaximus: My first iteration was actually this with implicit fallback logic, but I created the middleware creator thunk for a few reasons:

1. I think that it is quicker to to reason about the flow of data at the route composition site when it's explicit what the middleware will use as its source of truth at the point where we create the middleware.
2. I think its safer and less likely to introduce subtle bugs as we add more routes if the person designing the route explicitly chooses which source to initialize the `ctx.sequelizeAccount` from, rather than having implicit fallback behavior that can result in the source of truth for the `sequelizeAccount` being different across individual requests to the same route depending on the content of the URL and the payload being sent.  In particular, I was thinking about any POST route with an optional `accountId` parameter, and how easy it would be to have the middleware target the wrong account for some requests in that case.  If we **never** have an optional `param` for `accountId` on any route then the the implicit fallback logic would be fine flexibility-wise, but I don't like relying on this kind of incidental behavior when there could well be a route like that in the future, which would require refactoring the implicit fallback to use this pattern anyway (I'd prefer to just touch this once middleware once).

If you have a strong opinion on this, I'll remove the interim middleware creator function entirely, but I think having the developer make _explicit_ choices about what data to use to initialize the sequelizeAccount is preferable, and the complexity cost here (like you mentioned, it's very common to have middleware creator thunks with arguments) is minimal.

app.js
MaximusHaximus: I see that the `koa-ratelimit` middleware shadowed the same named function for the actual middleware function as it did for the middleware creator, but I'm not convinced that it is a good pattern to mimic generally.  These two functions do two completely different things, and run at different points in the lifecycle of the app.  One is a synchronous function that is run only once, during app setup where the other is an async method that is executed with every request that comes in.  Also, a key difference between this and the `koa-ratelimit` implementation is that this thunk actually does validation synchronously, during creation of the middleware, so there is an expectation that an error can be  thrown from the middleware _creator_ function; I generally especially try to avoid shadowing functions with identical names in cases like this where errors can be thrown from either 'layer' of the thunk, even apart from the other issues I mentioned above

middleware/createAccount.js
MaximusHaximus: :+1: killing it now

middleware/fundedAccount.js
```diff
@@ -88,9 +70,15 @@ const createFundedAccount = async (ctx) => {
         return;
     }
 
-    const fundingAccount = await ctx.near.account(fundedCreatorKeyJson.account_id);
 
-    // TODO: Should the client get something different than the result of this call?
+    const [sequelizeAccount, fundingAccount] = Promise.all([
```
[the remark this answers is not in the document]
MaximusHaximus: :+1: good catch; this would only fail in the error case. I'll add a test for this case also :+1:

middleware/fundedAccount.js
MaximusHaximus: :+1: good catch -- modified this to `403 Forbidden`

MaximusHaximus:
@vgrichina Resolved the blockers and responded to comments.  I've got a note for myself to add test coverage for the `funded account` routes.  Unfortunately the way our middleware modules depend directly on `process.env` makes it impossible to test both the success and failure cases, so adding proper coverage for both cases will require adding some encapsulation around the middleware composition that I'd like to do carefully and as its own changeset, without blocking landing this PR :)

PR #447, opened 2021-06-17
Allow concurrent creation of new accounts for testnet

MaximusHaximus:
Adds MultiKeyAccountCreator class that is responsible for using multiple keys to try to create new accounts, in a round-robin style.  Retrying due to a failed attempt to create an account will use the next key in sequence each time we retry, making it more likely that a retry will actually succeed when we are under heavy account creation load.

Our `near-api-js` keystore implementation and internals don't really provide any way to use multiple FullAccess keys to submit transactions; there is a low level assumption that there will be one-and-only-one key to be used for transactions.  As a result, I imported chunks of `near-api-js` and wrote our own code to handle the round-robin case.  Basically, this bypasses the Account object's `createAccount()` method and `signAndSendTransaction()` chain of calls entirely so that we can control the key being used for the signing of the transactions.

Because I bypassed the entire `signAndSendTransaction()` call chain, I also implemented a basic function to handle retrying with jitter.


MaximusHaximus:
Merging this because testnet account creation is unusable without this PR (thus `dev deploy` is also broken)

PR #449, opened 2021-06-17
Release 2021-06-17

MaximusHaximus:
1. Enables accurate detection of user IP addresses for our rate limiting middleware by respecting proxy server `X-Forwarded...` headers provided by render
2. Includes round-robin account creation _just on the `/account` route that is used on `testnet`_; no impact for `mainnet.
3. Added `debug` package so we can have some debug logging that is disabled unless we toggle it in by setting environment variable `DEBUG` to include the module(s) we want debug logging for
4. Tweaked `/fundedAccount` so that if someone hits that route when there is no access key configured, that it gracefully returns `available: false`
5. Added `/likelyNFTs` route; prototype implementation


MaximusHaximus:
Merging this to unblock Patrick's MVP NFT tokens implementation work

Patrick1904 (approved):

PR #474, opened 2021-07-27
Add `/fiat` endpoint

### LINKED ISSUE #473: Add support to provide fiat currency conversion for NEAR tokens
Part of https://github.com/near/near-wallet/issues/1854

1. New route: `/fiat`
2. Route returns an object hash containing current fiat conversion for NEAR->USD

- Fetch Fiat conversion from CoinGecko (https://www.coingecko.com/en/api)
- Return object hash with shape `{ NEAR: 2.140 }` to faciliate returning other tokens' fiat conversions in the future without changing API endpoint response shape
- Cache value in memory; only hit the CoinGecko API every 10s (CoinGecko's API throttles at 50 requests per minute; since we run 3 contract helper instances for availability/scalability reasons, we need to be sure that the total # of requests per minute for all 3 is well under 50.

MaximusHaximus:
Closes #473 

**Adds a new endpoint: `/fiat`**
- Endpoint returns `USD`, `EUR` and `CNY` fiat values for NEAR tokens.
- Endpoint matches CoinGecko's API shape for returned data.
- Using `node-cache` to provide a 10s cache period for data fetched from the CoinGecko API
- Using `dataloader` to provide batching and coalescing of requests for identical data from multiple users.

Example output:
```javascript
{
    "near": {
        "usd": 2.08,
        "eur": 1.76,
        "cny": 13.49,
        "last_updated_at": 1627351047
    }
}
```

- Endpoint can be easily enhanced in the future to return fiat values of tokens other than NEAR, against a wider variety of fiat currencies.


Patrick1904 (approved):
LGTM 👍

marcinbodnar (approved):

esaminu (approved):

Issue #524, opened 2021-09-28
Add structured logging for key routes using Bunyan


MaximusHaximus:
- Bunyan for logging
- Stream implementation for Sumologic (https://github.com/TonyFNZ/bunyan-sumologic)
- Add request level logging to routes with side-effects of note (SMS for now) including key data we need for fraud detection

PR #525, opened 2021-09-29
Request logging middleware

andy-haynes:
This PR adds new middleware for logging request metadata and payload information via Bunyan to SumoLogic. The middleware is added to endpoints responsible for sending SMS messages.

Addresses #524


MaximusHaximus (changes requested):
Just two things, then we can get this sucker landed

PR #529, opened 2021-10-12
Mieubrisse/fix dockerfiles

mieubrisse:


.gitpod.yml
```diff
@@ -1,5 +1,6 @@
 image: gitpod/workspace-postgres
 tasks:
+  # NOTE: As of 2021-1012, the contract helper DB seems to use the indexer DB so this likely won't work anymore
   - init: yarn && ./scripts/create_dev_dbs.sh
```
MaximusHaximus, line 4: It seems like this PR deletes this file, but still references it here in the `init` clause... can you confirm that we don't need `create_dev_dbs.sh` in the gitpod.yml file? AFAIK it is necessary to configure the Postgres gitpod instance.

scripts/create_dev_dbs.sh
```diff
@@ -1,9 +0,0 @@
-#!/bin/sh
```
MaximusHaximus, line 1: I think removing this file will break gitpod support; was it not working in the first place? Just want to make sure we're not removing existing functionality :)

MaximusHaximus (changes requested):
Just a few questions in-line -- thanks for the contribution, we really appreciate your time :)

PR #530, opened 2021-10-19
Model endpoint decoupling

andy-haynes:
This PR refactors database queries and mutations across the codebase to encapsulate the data layer implementation within behavioral logic. This change effectively decouples `sequelize` from the endpoint logic, establishing an abstraction layer through which data can be accessed and changed.

The behavioral logic lives under the `/services` directory, which contains modules for each database model. Methods on these service modules, which return only [POJOs](https://masteringjs.io/tutorials/fundamentals/pojo), represent the possible interactions with each model (e.g. creates, updates, fetches). Currently these methods call the `sequelize` methods created in `/services/sequelize`, which contains the `sequelize` logic refactored out of the endpoint code. This consolidation of the `sequelize` logic greatly simplifies the upcoming migration to another data source while introducing minimal changes in the existing codebase.


config/features.js
```diff
@@ -0,0 +1,30 @@
+const { ENVIRONMENTS } = require('./constants');
```
andy-haynes, line 1: This is a placeholder implementation of the accessor in the [`flag-editor` package](https://github.com/near/near-wallet/pull/1771) which will be imported directly when available. In the interim this implementation is subject to change.

services/sequelize/recovery_method.js
```diff
@@ -0,0 +1,169 @@
...
+        };
+    },
+
+    async createRecoveryMethod({ accountId, detail, kind, publicKey, requestId, securityCode }) {
+        const account = await Account.findOne({ where: { accountId } });
+        const recoveryMethod = await account.createRecoveryMethod({
+            kind,
+            ...(detail && { detail }),
+            ...(publicKey && { publicKey }),
+            ...(requestId && { requestId }),
```
MaximusHaximus, line 40: Although we store this as a string in our SQL DB currently, it is a number in the smart contract so we can assume it will be POST'd to us as a number.  Given that fact, if `requestId` is 0, this will be falsy and not included in the payload, but that is a valid index for `requestId` and should be posted as an attribute here AFAICT

services/sequelize/recovery_method.js
```diff
@@ -0,0 +1,169 @@
...
+        return this.cleanRecoveryMethod(recoveryMethod);
+    },
+
+    async deleteOtherRecoveryMethods({ accountId, detail }) {
+        const account = await Account.findOne({ where: { accountId } });
+        if (!account) {
+            return null;
+        }
+
+        const allRecoveryMethods = await account.getRecoveryMethods();
```
MaximusHaximus, line 53: When we port to DDB, we should make these deletes run in parallel :)

services/sequelize/recovery_method.js
```diff
@@ -0,0 +1,169 @@
...
+
+        return recoveryMethod.destroy();
+    },
+
+    async getTwoFactorRecoveryMethod(accountId) {
+        const twoFactorRecoveryMethod = await this.getTwoFactorRecoveryMethod_internal(accountId);
+        return this.cleanRecoveryMethod(twoFactorRecoveryMethod);
+    },
+
+    async getTwoFactorRecoveryMethod_internal(accountId) {
```
MaximusHaximus, line 78: Just thinking out loud -- node >= 14 supports private class methods and Render uses node v14.17...if we continue using it we could just mark this as private.  Not a blocker, though, we can come back to it.

services/sequelize/recovery_method.js
```diff
@@ -0,0 +1,169 @@
...
+    },
+
+    async listRecoveryMethods_internal({ accountId, detail, kind, publicKey, securityCode }) {
+        const account = await Account.findOne({ where: { accountId } });
+        const methods = await account.getRecoveryMethods({
+            where: {
+                ...(detail && { detail }),
+                ...(kind && { kind }),
+                ...(publicKey && { publicKey }),
+                ...(securityCode && { securityCode }),
```
MaximusHaximus, line 119: Same as above with securityCode === 0 being falsy but a valid value here

app.js
```diff
@@ -483,16 +587,95 @@ const completeRecoveryValidation = ({ isNew } = {}) => async ctx => {
...
+    if (!recoveryMethod) {
+        ctx.throw(401, 'recoveryMethod does not exist');
+    }
+
+    // for new accounts, clear all other recovery methods that may have been created
+    if (isNew) {
+        await RecoveryMethodService.deleteOtherRecoveryMethods({ accountId, detail: method.detail });
+    }
+
+    await RecoveryMethodService.updateRecoveryMethod({
```
MaximusHaximus, line 625: Looks like the update() call in the legacy function only set the one value (securityCode)... is providing the detail/kind here redundant or could this in some way change our existing behaviour?

services/sequelize/recovery_method.js
```diff
@@ -0,0 +1,169 @@
...
+        await recoveryMethod.update({ securityCode });
+        return this.cleanRecoveryMethod(recoveryMethod);
+    },
+
+    async updateRecoveryMethod({ accountId, detail, kind, securityCode }) {
+        const [recoveryMethod] = await this.listRecoveryMethods_internal({ accountId, detail, kind, securityCode });
+        await recoveryMethod.update({
+            ...(detail && { detail }),
+            ...(kind && { kind }),
+            ...(securityCode && { securityCode }),
```
MaximusHaximus, line 150: Same as above with securityCode === 0 being falsy but a valid value here

services/sequelize/recovery_method.js
```diff
@@ -0,0 +1,169 @@
...
+        return this.cleanRecoveryMethod(recoveryMethod);
+    },
+
+    async updateTwoFactorRecoveryMethod({ accountId, detail, kind, requestId, securityCode }) {
+        const twoFactorRecoveryMethod = await this.getTwoFactorRecoveryMethod_internal(accountId);
+        await twoFactorRecoveryMethod.update({
+            requestId,
+            ...(detail && { detail }),
+            ...(kind && { kind }),
+            ...(securityCode && { securityCode }),
```
MaximusHaximus, line 162: Same as above with securityCode === 0 being falsy but a valid value here

test/2fa.test.js
```diff
@@ -56,7 +56,7 @@ describe('2fa method management', function () {
             request,
         });
 
-        await models.sequelize.sync({ force: true });
+        await deleteAllRows();
```
MaximusHaximus, line 59: <3 bugfixes FTW

test/db.js
```diff
@@ -0,0 +1,10 @@
+const models = require('../models');
+
+module.exports = {
+    async deleteAllRows() {
```
MaximusHaximus, line 4: Not a blocker, but could we run these in parallel?

test/app.test.js
```diff
@@ -71,7 +93,7 @@ describe('app routes', function () {
             request,
         });
 
-        await models.sequelize.sync({ force: true });
+        await deleteAllRows();
```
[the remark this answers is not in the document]
MaximusHaximus: Good catch :+1:

MaximusHaximus (changes requested):
Added a few questions/concerns in-line.  Nice work on adding some clean interface heres -- this is a bit of a monster that has had a lot of functionality 'bolted on' without a lot of work on the underlying structure of the code 💖 .  Also, nice work on the test coverage for the services :+1:

features/environments.json
```diff
@@ -0,0 +1,7 @@
+{
```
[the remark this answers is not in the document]
andy-haynes: Nope I think this was a placeholder I forgot to remove, I don't see it referenced anywhere.

test/db.js
andy-haynes: I believe the first three can, `Accounts` has to be deleted last because there are foreign keys referencing `id`

middleware/2fa.js
```diff
@@ -234,100 +230,109 @@ const initCode = async (ctx) => {
...
-        await twoFactorMethod.update({
-            kind: method.kind,
+
+        await RecoveryMethodService.updateTwoFactorRecoveryMethod({
+            accountId,
             // FIXME: Should this be escaped?
-            detail: method.detail,
-            requestId: -1
+            detail,
+            kind,
```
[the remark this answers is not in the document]
andy-haynes: You are indeed, good catch!

middleware/2fa.js
```diff
@@ -197,21 +199,15 @@ const isContractDeployed = async (accountId) => {
     return MULTISIG_CONTRACT_HASHES.includes(state.code_hash);
 };
 
-const getAccountAndMethod = async (ctx, accountId) => {
-    const [account] = await models.Account.findOrCreate({ where: { accountId } });
+const getTwoFactorRecoveryMethod = async (ctx, accountId) => {
+    const account = await AccountService.createAccount(accountId);
     if (!account) {
```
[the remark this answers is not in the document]
andy-haynes: Auto-creating made it really hard to reason about the expected behavior in a lot of places 😕

services/sequelize/recovery_method.js
andy-haynes: 😎

services/sequelize/recovery_method.js
andy-haynes: You'd still have collisions with the public method, no? Anyway in this context I used `_internal` to indicate methods that return ORM objects as opposed to POJOs - I only wanted to return POJOs in the service objects so I moved all that jank under `/sequelize` 😎

app.js
andy-haynes: This is one of the things that makes this conversion so hard to reason about 🙁 

The fetch for `recoveryMethod` in the legacy method is:
```
const [recoveryMethod] = await account.getRecoveryMethods({
    where: {
        kind: method.kind,
        detail: method.detail,
        securityCode: securityCode
    }
});

...

await recoveryMethod.update({ securityCode: null });
```

Since `recoveryMethod` necessarily has `detail === method.detail` and `kind === method.kind` already, updating _should_ be a no-op. Then in `/sequelize/recovery_method.js`, the `updateRecoveryMethod` method does the same fetch anyway:
```
async updateRecoveryMethod({ accountId, detail, kind, securityCode }) {
    const [recoveryMethod] = await this.listRecoveryMethods_internal({ accountId, detail, kind, securityCode });
    ...
}
```

which ensures that it's still fetching the same thing even if it is doing a redundant update. It's possible that the fetch is nondeterministic if another `detail` + `kind` pair existed with a different `publicKey` (the combination of which constitutes a unique tuple based on the index), but since it just takes the first value in the destructure anyway, the logic should be identical.

It's still a bug however, since I'm not actually setting `securityCode = null` on line 625 🤦. So you might want to double check my explanation once more to make sure it parses 😄

services/sequelize/recovery_method.js
MaximusHaximus: True on the method name conflict, 🤦  :)
Roger that on the _internal naming due to the rich object vs. POJO :+1: I can dig it.

MaximusHaximus (approved):
Epic work, nice job <3.  Lets get this sucker landed! :D

PR #532, opened 2021-11-16
feat: add query to indexer to collect mint events for accountId

willemneal:
The new table `assets__non_fungible_token_events` contains logged nft events.

Currently this query will find all distinct contracts that emitted an event involving the accountId.  In the future this could be improved, e.g. tracking transfers and burns to exclude no longer valid contract associations.

### DIFF
--- a/middleware/indexer.js
+++ b/middleware/indexer.js
@@ -151,7 +151,7 @@ const findLikelyTokens = async (ctx) => {
 const findLikelyNFTs = async (ctx) => {
     const { accountId } = ctx.params;
 
-    const received = `
+    const ownershipChangeFunctionCalls = `
         select distinct receipt_receiver_account_id as receiver_account_id
         from action_receipt_actions
         where args->'args_json'->>'receiver_id' = $1
@@ -160,13 +160,13 @@ const findLikelyNFTs = async (ctx) => {
             and args->>'method_name' like 'nft_%'
     `;
 
-    const mintEvents = `
+    const ownershipChangeEvents = `
         select distinct emitted_by_contract_account_id as receiver_account_id 
         from assets__non_fungible_token_events
         where token_new_owner_account_id = $1
     `;
 
-    const { rows } = await pool.query([received, mintEvents].join(' union '), [accountId]);
+    const { rows } = await pool.query([ownershipChangeFunctionCalls, ownershipChangeEvents].join(' union '), [accountId]);
     ctx.body = rows.map(({ receiver_account_id }) => receiver_account_id);
 };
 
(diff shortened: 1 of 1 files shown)


willemneal:
@MaximusHaximus I'm still not sure how to ensure that this is updated to use the newly released indexer for testnet.

willemneal:
@MaximusHaximus Anything I can help with?

frol (approved):
This is just the first step of the NFT events support. Currently, we cannot completely switch to `assets__non_fungible_tokens_events` table since it does not contain records about the NFT "events" before NFT Events NEP.

I confirm that this PR is an improvement already since it will bring the list of all the NFT contracts that minted/burned/transferred NFT to a given account.

frol:
@MaximusHaximus Using the `assets__non_fungible_tokens_events` table you can be more flexible and even fetch a history of the changes.

MaximusHaximus (changes requested):
:+1: Thanks for the submission, @willemneal, we really appreciate it!  Sorry about the delayed response, it's been a bit hectic recently.

I have two small requests, otherwise this is GTG:

1. One detail is that this query will also find TRANSFER events, even if they were generated from a batch transfer where someone didn't call an `nft_xxxxxxx` method name -- which is a good thing, since previously these types of transfers were invisible to the wallet frontend.

2. For clarity in code around this behaviour, can we re-name `mintEvents` to something that indicates that the result set returned doesn't _only_ include mint events?  Something like `ownershipChangeEvents`.  Once we've done that, it'd be nice to also rename the existing `received` variable to match the level of detail of the new variable also -- maybe something like `ownershipChangeFunctionCalls` to provide immediate clarity that we are combining two different sources of truth here without the user needing to read + understand the SQL schemas involved?

Thanks again for the submission!

willemneal:
@MaximusHaximus No problem!  I addressed your naming issues but still have one question:

> 1. One detail is that this query will also find TRANSFER events, even if they were generated from a batch transfer where someone didn't call an nft_xxxxxxx method name -- which is a good thing, since previously these types of transfers were invisible to the wallet frontend.

Is this a request or just prefacing why the name was bad?

MaximusHaximus:
> @MaximusHaximus No problem! I addressed your naming issues but still have one question:
> 
> > 1. One detail is that this query will also find TRANSFER events, even if they were generated from a batch transfer where someone didn't call an nft_xxxxxxx method name -- which is a good thing, since previously these types of transfers were invisible to the wallet frontend.
> 
> Is this a request or just prefacing why the name was bad?

Just prefacing why the name could be clearer, no request there :)

PR #545, opened 2022-01-19
fix: add testnet stake&farm factory pool pattern

esaminu:
Adds factory from [staking-farm](https://github.com/referencedev/staking-farm/blob/master/HowTo.md) repo to indexer staking pool queries on testnet.


MaximusHaximus (approved):
:+1:

PR #551, opened 2022-02-04
WAL-90 chore: Remove USE_DB_SERVICES flag

andy-haynes:


test/app.test.js
```diff
@@ -237,11 +220,7 @@ describe('app routes', function () {
         it('returns 403 Forbidden (signature not from accountId owner)', async () => {
             // FIXME: This is just testing incorrect blockNumber, *not* that the signature is from a different owner
             const accountId = await testAccountHelper.createNEARAccount();
-            if (!USE_DB_SERVICES) {
```
morgsmccauley, line 240: In future we might want to think about creating/duplicating a whole new test targeted against the flag on state, leaving the original flag off test as is. This way both flag states will be tested meaning we can catch potential issues coming from changes within each code branch.

andy-haynes: That's a really good point, changing the behavior inline like this is an _excellent_ way to introduce bugs that pass tests 😉. My main concern, and the reason I generally flag inline for less complex tests, is that I find duplicating makes it easier to introduce subtle differences in tests which, due to to the amount of duplicated code added, can be harder to detect than changes made within individual tests. It also makes it more difficult to reference multiple feature flags in a single test, which admittedly shouldn't come up in ideal circumstances. I still think duplicating the test is a good idea when the logic has been refactored more extensively (e.g. the deleted code at line 469) but I tend to use it only when inline checks would make it harder to reason about the changes.

However the ultimate goal with the CI/CD pipeline will be to run tests across all environments in parallel. So instead of considering individual flag states, we will be testing the set of flag states that exist across environments. The progression of a feature flag's enablement will ensure that tests evaluating both flag states will be run simultaneously in the pipeline, at least until the flag is enabled in every environment. To me this achieves something functionally equivalent - until a flag is enabled in the final production environment, test suites are run with the flag both on and off.

morgsmccauley: Good points, running against every point would make the duplication unnecessary and is a much cleaner approach 🎉.

PR #552, opened 2022-02-05
WAL-94 feat: Jenkins AWS deployment

andy-haynes:
This PR adds a Jenkinsfile for deploying to the new AWS environments. The container defined in the new `Dockerfile` is built and pushed to ECR before updating the staging and live services running on the target environment's ECS cluster.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "app.js",
   "scripts": {
     "start": "env $(sed 's/ # .*//' .env) supervisor app",
+    "docker:start": "node app.js",
     "test": "yarn lint && mocha",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "app.js",
   "scripts": {
     "start": "env $(sed 's/ # .*//' .env) supervisor app",
+    "docker:start": "node app.js",
```
andy-haynes, line 8: This is referenced by the Terraformed ECS services' container definitions.

MaximusHaximus (approved):
:+1: nice work, excited to see this pipeline _active_ :)

PR #559, opened 2022-02-16
feat(2fa): don't bypass initial 2FA verification for existing recovery methods

### LINKED ISSUE #558: No email check after re-enabling 2FA on the same address
**Problem**
As a user, I want to receive a confirmation email (with the six digits code) when I enable 2FA.
Currently, re-enabling the 2FA on the same account using the same email won't send the confirmation email.

**Expected Behavior**
Always send a confirmation email when 2FA is enabled, even after the first time.

**Steps to reproduce**

1. enable 2FA via email on an account
2. confirm the activation with the six digits code
3. issue a transaction to test if the 2FA is enabled
4. disable 2FA from the account page
5. re-enable 2FA using the same email

andy-haynes:
When a user signs up for 2FA with an email address/SMS number that is already used as a recovery method, the current behavior is to not send the initial verification code as the communication channel has already been verified by virtue of belonging to an existing recovery method owned by the account holder. However this behavior can be unexpected and give the impression to users that their 2FA contract is not securely configured, undermining any convenience that this behavior may have sought to provide.

By ceasing to query for a corresponding recovery method, we ensure that the enabling 2FA always sends an initial code for the user to verify.

Closes #558

### DIFF
--- a/middleware/2fa.js
+++ b/middleware/2fa.js
@@ -263,21 +263,6 @@ const initCode = async (ctx) => {
         });
     }
 
-    // check if 2fa method matches existing recovery method
-    const [recoveryMethod] = await recoveryMethodService.listRecoveryMethods({
-        accountId,
-        detail,
-        kind: kind.split('2fa-')[1],
-    });
-
-    if (recoveryMethod) {
-        // client should deploy contract
-        ctx.body = {
-            confirmed: true, message: '2fa initialized and set up using recovery method verification'
-        };
-        return;
-    }
-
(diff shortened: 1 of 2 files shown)


andy-haynes:
Initially I had a feature flag around this functionality but given the low impact of this change and the pending platform migration I think it's simpler to not use a flag and simply verify the behavior in testnet.

Patrick1904 (approved):
Sweet, looks good! 👍

PR #562, opened 2022-02-24
chore: add new factory to pool pattern

esaminu:
Add validators using wallet's factory `factory.colorpalette.testnet` as well as aurora's `factory01.littlefarm.testnet` to pool pattern


MaximusHaximus (approved):
:+1:

PR #567, opened 2022-02-28
WAL-115 Disable initialisation of phone based 2FA

morgsmccauley:
This PR disables the initialisation of 2FA via SMS in preparation for its sunsetting.

Code verification is still needed when disabling a 2FA via SMS which is why the majority of the related code still remains. This PR simply short-circuits the creation of any new methods.

⚠️ This should not be landed until the relevant UI has been removed from NEAR wallet frontend.


test/2fa.test.js
```diff
@@ -92,9 +101,6 @@ describe('2fa method management', function () {
     });
 
     describe('changing an already requested 2fa method', () => {
```
andy-haynes, line 103: Does this test do anything but change the `kind` of the recovery method? If it's no longer providing any value now that we prevent changing `kind` then it's probably not worth keeping around.

morgsmccauley: The important part here is the `should allow verification using the code generated by the second request` test inside. This will ensure that a user started the 2FA setup but didn't verify the sent code, can still start the process again and successfully verify with the new method (starting the process again will update the existing `RecoveryMethod`).

I think this is still relevant even though we are no longer updating the `kind`, but I think the `describe` message should be updated to something more relevant 🤔 maybe `completing an already requested 2fa method`?

Thoughts?

andy-haynes: Good point, I thought that was covered separately but I may be getting confused with an unmerged PR. In that case I think changing the name is the right way to go 👍

andy-haynes (approved):
Nice work

PR #571, opened 2022-03-16
WAL-160 DynamoDB fixes

andy-haynes:
This PR addresses issues outlined in [WAL-160](https://nearinc.atlassian.net/browse/WAL-160):
- adds an update expression to the recovery method `update` DB method to ensure a value for `createdAt` when document is created via upsert
- use `publicKey` in recovery method validation endpoints to ensure the full range key can be composed (requires [near-wallet PR](https://github.com/near/near-wallet/pull/2529))
- update service calls to always use the DB methods injected in the service constructor, not the DB method imported into module scope
- restore the missing `kind` parameter to `getIdentityVerificationMethod` call when querying Postgres (removed in #550)

### DIFF
--- a/db/methods/recovery_method/update.js
+++ b/db/methods/recovery_method/update.js
@@ -11,6 +11,9 @@ function updateRecoveryMethod({ accountId, kind, publicKey }, { detail, requestI
         publicKey,
         requestId,
         securityCode,
+    }, {
+        UpdateExpression: 'SET createdAt = if_not_exists(createdAt, :now)',
+        ExpressionAttributeValues: { ':now': (new Date()).toISOString() },
     });
 }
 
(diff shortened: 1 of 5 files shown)


db/methods/recovery_method/update.js
```diff
@@ -11,6 +11,9 @@ function updateRecoveryMethod({ accountId, kind, publicKey }, { detail, requestI
         publicKey,
         requestId,
         securityCode,
+    }, {
+        UpdateExpression: 'SET createdAt = if_not_exists(createdAt, :now)',
+        ExpressionAttributeValues: { ':now': (new Date()).toISOString() },
```
morgsmccauley, line 16: 🤩

morgsmccauley (approved):

PR #574, opened 2022-03-22
WAL-178 Use Twilio Verify for 2FA phone verification

morgsmccauley:
This PR migrates 2FA phone verification from to [Twilio Verify](https://www.twilio.com/verify). Verify handles delivery of the OTP as well as verification. Messages will now be along the lines of:
```
Your NEAR Wallet verification code is: XXXXXX
```
Contrast to the existing SMS messages this no longer contains the transaction details of the 2FA request. This has now been replaced by additional UI in the frontend merged in this [PR](https://github.com/near/near-wallet/pull/2523).

This **only** affects SMS based 2FA methods. Email and SMS recovery methods (now disabled) still use the existing custom solution.

This PR also adds a stubbed implementation of the Verify API so that it can be bypassed in local development and test. This has been put behind the `USE_MOCK_TWILIO` environment variable. This is off be default by will be enabled for `development` and `test`.


services/twilio_verify.js
```diff
@@ -0,0 +1,51 @@
+const twilio = require('twilio');
+const password = require('secure-random-password');
+
+const ACCOUNT_SID = process.env.TWILIO_ACCOUNT_SID;
+const AUTH_TOKEN = process.env.TWILIO_AUTH_TOKEN;
+const VERIFY_SERVICE_SID = process.env.TWILIO_VERIFY_SERVICE_SID;
```
morgsmccauley, line 6: ⚠️ Need to add this to `testnet` and `mainnet`

test/2fa.test.js
```diff
@@ -202,94 +193,163 @@ describe('2fa method management', function () {
...
+                    })
+                        .then(expectInvalid2faCodeProvidedError);
+                });
             });
         });
 
-        describe('malformed security codes should be rejected', () => {
-            const expectInvalid2faCodeProvidedError = expectFailedWithCode(401, 'invalid 2fa code provided');
-            const expectInvalid2faCodeForRequestError = expectFailedWithCode(401, '2fa code not valid for request id');
+        describe('email', () => {
```
morgsmccauley, line 280: Above changes are just rearrangements, real changes are from here ⬇️ . Now that `2fa-email` and `2fa-phone` have different verification paths i've added tests to cover both.

app.js
```diff
@@ -524,6 +524,12 @@ router.post('/account/validateSecurityCodeForTempAccount',
 const createFiatValueMiddleware = require('./middleware/fiat');
 router.get('/fiat', createFiatValueMiddleware());
 
+if (process.env.NODE_ENV === 'development') {
+    app.on(SERVER_EVENTS.SECURITY_CODE, (code) => {
+        console.log('Security code: ', code);
```
morgsmccauley, line 529: Logging security codes to the console so it's still possible to verify a `2fa-phone` method without Twilio Verify

middleware/2fa.js
```diff
@@ -234,7 +250,7 @@ const initCode = async (ctx) => {
     }
 
     const { kind, detail } = method;
-    if (kind !== TWO_FACTOR_AUTH_KINDS.EMAIL) {
+    if (process.env.NODE_ENV !== 'test' && kind !== TWO_FACTOR_AUTH_KINDS.EMAIL) {
```
morgsmccauley, line 253: We need to be able to create SMS 2FA methods in `test` so we can test verification

test/2fa.test.js
```diff
@@ -89,15 +89,6 @@ describe('2fa method management', function () {
             }
         });
 
-        it('prevents initialisation of phone based 2FA methods', async () => {
```
morgsmccauley, line 92: Removing as we now need this flow in `test`

services/twilio_verify.js
```diff
@@ -0,0 +1,51 @@
...
+const USE_MOCK_TWILIO = process.env.USE_MOCK_TWILIO === 'true';
+
+const VERIFY_DELIVERY_CHANNEL = 'sms';
+
+module.exports = class TwilioVerifyService {
```
andy-haynes, line 11: Do you think there would be any value in having separate classes for mock vs live instances? This is kind of an unusual case in that the interface is simple but every method needs to check whether it should be mocking and effectively does something completely different.

Curious to hear your thoughts as to whether the overhead of a "duplicate" class implementing the same interface would be a good tradeoff here. This is something that might be more clear in TypeScript with actual `interface`s.

test/2fa.test.js
```diff
@@ -202,94 +193,163 @@ describe('2fa method management', function () {
     });
 
     describe('/verify', function () {
-        let accountId;
-        let securityCode;
+        const invertSecurityCode = (code) => [...code].map((digit) => 9 - digit).join('');
```
andy-haynes, line 196: 😎

services/twilio_verify.js
```diff
@@ -0,0 +1,51 @@
+const twilio = require('twilio');
```
andy-haynes, line 1: Do any of the `twilio` methods called here throw exceptions we need to wrap before returning REST responses? If so it might be worthwhile to pass `twilio` as a constructor parameter so that we can mock failures in the test suite.

services/twilio_verify.js
andy-haynes: Will we be using the same values we currently have for `TWILIO_ACCOUNT_SID` and `TWILIO_AUTH_TOKEN` or does Twilio Verify use different configuration?

andy-haynes (changes requested):
Looking good, minor feedback inline

middleware/2fa.js
```diff
@@ -286,19 +302,39 @@ const sendNewCode = async (ctx) => {
     };
 };
 
+const validateCodeByKind = async (ctx, twoFactorMethod, code) => {
+    switch (twoFactorMethod.kind) {
+    case TWO_FACTOR_AUTH_KINDS.PHONE:
```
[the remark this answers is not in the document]
morgsmccauley: This is what `eslint` wants 🥲
image

middleware/2fa.js
```diff
@@ -286,19 +302,39 @@ const sendNewCode = async (ctx) => {
...
+        return twilioVerifyService.verify({ to: twoFactorMethod.detail, code });
+    case TWO_FACTOR_AUTH_KINDS.EMAIL:
+        return twoFactorMethod.securityCode === code;
+    default:
+        break;
```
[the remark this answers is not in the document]
morgsmccauley: This will because of the `return false;` on L317, do you think it would be more explicit to add it here too?

middleware/2fa.js
```diff
@@ -286,19 +302,39 @@ const sendNewCode = async (ctx) => {
     };
 };
 
+const validateCodeByKind = async (ctx, twoFactorMethod, code) => {
```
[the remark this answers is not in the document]
morgsmccauley: 😅 👍🏽

services/twilio_verify.js
morgsmccauley: I thought about this for a while too and do think a separate class is the ideal case but we don't have a good way of injecting or overriding the live instance when we'd like. Two approaches I thought of were:

1. Using Koa context:
```js
// in app.js

app.context.services = {
  twilioVerify: new TwilioVerifyService(),
} ;

// later somewhere, also need to do it for `test`
if (process.env.NODE_ENV === 'development') {
  app.context.services = {
    twilioVerify: new TwilioVerifyMockService(),
  } ;
}
```

2. Conditional `require`
```js
// in app.js

const TwilioVerifyService = process.env.NODE_ENV === 'development' ? require('mock') : require('real');
```

2 wouldn't work for `test` since it's hardcoded in module scope but 1. Could. It then became more confusing because the live and mock classes need different arguments:
```js
// live
send({ to }) {
    return this.verifyService.verifications.create({ to, channel: VERIFY_DELIVERY_CHANNEL });
}

// mock
send({ to }, emitServerEvent) {
    const securityCode = this.generateMockSecurityCode();

    this.securityCodes[to] = securityCode;
    emitServerEvent(securityCode);
}
```
So we'd always need to pass in `emitServerEvent` but we'd actually only be used _if_ we override the live instance with the mock.

This is sort of why I was more inclined to merge the two together but i'm still on the fence, keen to hear your thoughts on the above!

services/twilio_verify.js
morgsmccauley: Yes same values. Twilio Verify is a sub service of the main account.

middleware/2fa.js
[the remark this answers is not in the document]
andy-haynes: You wouldn't need the `return` on L317 because the `default` is guaranteed to be hit. `switch` is weird 🙂

app.js
morgsmccauley: Oh yeah good catch, it's actually an object 🤦🏽‍♂️ . I'll fix this up.

middleware/2fa.js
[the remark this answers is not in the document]
andy-haynes: Madness! I've never seen it not indented. There is an override for it but it's not really a big deal. Just looks jarring to see non-indented code inside a block.

middleware/2fa.js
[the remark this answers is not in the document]
morgsmccauley: True 😄  i'll update!

middleware/2fa.js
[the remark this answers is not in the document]
morgsmccauley: Yeah I actually hate these, i've updated `eslint` to allow this 😄

services/twilio_verify.js
andy-haynes: Hmm good points, I like the `context` override solution but it may not be very durable if we move off of Koa later. Another non-invasive way might be something like:
```
module.exports = {
    TwilioService: useMockService ? MockService : TwilioService,
};
```
but that introduces a small inconsistency with the services that use a default export. The `emitServerEvent` is trickier but maybe could be worked into the mock's constructor... still not great though I think.

In this case it might be easier to just rely on the simplicity of the interface, it's not like it's particularly hard to reason about as is 😄

middleware/2fa.js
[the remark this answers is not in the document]
andy-haynes: 😍

services/twilio_verify.js
morgsmccauley: I think we can do:
```js
module.exports = USE_TWILIO_MOCK ? TwilioServiceMock : TwilioService;
```
Still have the issue of arguments though 😅

morgsmccauley: Starting to think that's a cleaner option ⬆️  should create better separation and with both classes in the same file the arguments issue shouldn't be too complicated to figure out 😄  i'll push up a change.

services/twilio_verify.js
morgsmccauley: None worth handling, only specific errors for invalid/missing arguments:
https://github.com/twilio/twilio-node/blob/main/lib/rest/verify/v2/service/verificationCheck.js#L76-L81
https://github.com/twilio/twilio-node/blob/main/lib/rest/verify/v2/service/verification.js#L93-L101

morgsmccauley: Actually i'm wrong - looks like there is a max attempts error that we should cover, and probably more. I'll have a deeper look in to this 👀

morgsmccauley: Updated to handle errors we're likely to encounter, and as you suggested passing `twilio` as an argument so it could be tested :)

andy-haynes (changes requested):
This is awesome ❤️ 

Just a couple questions on error handling but it's looking great.

services/twilio_verify/twilio_verify.js
```diff
@@ -0,0 +1,52 @@
...
+        twilio = twilioClient,
+    } = {}) {
+        const client = twilio(ACCOUNT_SID, AUTH_TOKEN);
+
+        this.verifyService = client
+            .verify
+            .services(VERIFY_SERVICE_SID);
+    }
+
+    handleTwilioErrors(fn) {
```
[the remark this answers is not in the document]
morgsmccauley: I think this works well with `send` which is a one liner, but not so much with `verify`:
```js
const { valid } = await this.verifyService.verificationChecks.create({ to, code })
    .catch((error) => {
        this.handleTwilioErrors(error);
    });
return valid;
```
I was sorta taking inspiration from `bluebird`s `Promise.try()` here, I like the idea of wrapping everything as it gives us the flexibility to compose many twilio methods and `catch` all errors which may be thrown from the individual methods. With this approach we'd need to apply to all methods individually or similarly wrap everything.

That was my reasoning but i'm on the fence, what are your thoughts? :)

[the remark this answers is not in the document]
morgsmccauley: With regards to stacktrace node still picks up the correct method names:
image

services/twilio_verify/twilio_verify.js
```diff
@@ -0,0 +1,52 @@
...
+const ACCOUNT_SID = process.env.TWILIO_ACCOUNT_SID;
+const AUTH_TOKEN = process.env.TWILIO_AUTH_TOKEN;
+const VERIFY_SERVICE_SID = process.env.TWILIO_VERIFY_SERVICE_SID;
+
+const VERIFY_DELIVERY_CHANNEL = 'sms';
+
+// https://www.twilio.com/docs/api/errors/XXXXX
+const ERROR_RESPONSES_BY_CODE = {
+    20404: { status: 429, text: '2FA request expired, please try again' },
+    20429: { status: 429, text: '2FA rate limit reached, please try again later' },
```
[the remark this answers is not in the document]
morgsmccauley: This one is a little confusing, the error message from Twilio is actually: `Rate limit for phone and service reached` which I assume means it is per user and not the service but still need to confirm this with two different devices.

500 definitely makes sense if it is service wide but 429 probably better is it is user based. I'll test the above and let you know the result :)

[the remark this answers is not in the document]
morgsmccauley: Tested the above and confirmed this is per phone. I updated the response message to make this a bit clearer :)

services/twilio_verify/twilio_verify.js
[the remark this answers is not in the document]
andy-haynes: For `verify` I think a `try`/`catch` would work pretty well 🙂 

I see your point about `Promise.try`, though it still seems a bit of a heavy abstraction in the current implementation since the two methods using it are only calling a single method capable of throwing a relevant error. Coupled with the fact that it's not doing anything more complex than augmenting the error it struck me as a bit unconventional.

But I think I'm getting more hung up on the syntax and low sample size, when I think of the behavior as more of a decorator (e.g. `@handleTwilioErrors()`) it seems like a reasonable way to encapsulate the common error logic in a composable way.

andy-haynes (approved):

services/twilio_verify/twilio_verify.js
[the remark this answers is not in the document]
morgsmccauley: > As long as the 429s we return are actionable by the client then I'm happy 👍

Definitely, that's what i'm working on now :)

morgsmccauley:
Closes #577

middleware/2fa.js
morgsmccauley: Re-enabled this and instead created/using a backdoor for creating 2fa phone methods. That way we can keep this disabled and have a test backing it which ensures it's disabled across all envs.

PR #575, opened 2022-03-22
test(db): update tests for DynamoDB integration

andy-haynes:
This PR addresses breaking tests when the `USE_DYNAMODB` feature flag is enabled, as well as adding test coverage for new methods and deprecating coverage for methods not called when `USE_DYNAMODB` is enabled.

These changes to are primarily due to minor disparities between how `sequelize` and `dynamodb` handle particular data interactions but also include one-off changes to existing tests for more consistent behavior and to accommodate the in-memory DynamoDB instance, including:
- validation of `null` input values
- casting of date types returned from the DB
- record creation behavior when a matching record already exists
- missing comparison of data points
- providing missing data to endpoints

### DIFF
--- a/test/services/identity_verification_method.js
+++ b/test/services/identity_verification_method.js
@@ -1,3 +1,5 @@
+require('dotenv').config({ path: 'test/.env.test' });
+
 const Constants = require('../../constants');
 const IdentityVerificationMethodService = require('../../services/identity_verification_method');
 const IdentityVerificationMethodSequelize = require('../../services/sequelize/identity_verification_method');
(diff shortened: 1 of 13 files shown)


test/services/identity_verification_method.js
```diff
@@ -1,3 +1,5 @@
+require('dotenv').config({ path: 'test/.env.test' });
```
morgsmccauley, line 1: Curious why this is necessary now?

morgsmccauley (approved):
LGTM + 1 question :)

test/services/identity_verification_method.js
andy-haynes: `dynamodb-local` requires values for `AWS_`* env vars, I probably could have overridden them inline as well but figured sourcing from env vars is closer to how it will function in a live environment.

PR #594, opened 2022-04-16
feat: add indexer query for fungible token events

esaminu:
Add indexer queries for any Fungible Token ownership change events per Fungible Token Event [standard](https://nomicon.io/Standards/Tokens/FungibleToken/Event)


MaximusHaximus (approved):
:+1: nice work! Thanks @esaminu

PR #599, opened 2022-04-26
feat(2fa): Multisig transfer script

andy-haynes:
This PR adds a script to transfer the communication channel for a multisig contract from a phone number for SMS messages to an email address. This is a stopgap measure to fix 2FA recovery methods for users who can no longer receive Twilio SMS messages and is only to be used once the user has confirmed their identity and ownership of the communication methods in question.

This script is invoked with command line arguments for the account ID, current phone number, and desired email address. Since the `kind` property, which is changed here, is part of the range key for `recovery_methods` documents, the document cannot be updated directly. A new document must be created with a `kind` value of `2fa-email` and the old document must be deleted.


scripts/multisig.js
```diff
@@ -0,0 +1,140 @@
...
+
+    if (process.env.NEAR_WALLET_ENV.startsWith('testnet')) {
+        return accountId.endsWith('.testnet');
+    }
+
+    return accountId.endsWith('.near');
+}
+
+function isEmail(detail) {
+    return detail.includes('@');
```
andy-haynes, line 28: This is overly simplistic but in this context we technically only care that it's not a phone number. Looking for the `@` just makes a convenient, dead simple validation on the email input.

scripts/multisig.js
```diff
@@ -0,0 +1,140 @@
...
+        console.warn('Aborting');
+        return;
+    }
+
+    await deleteRecoveryMethod({
+        accountId,
+        kind: currentMethodKind,
+    });
+
+    const email2faMethod = await updateRecoveryMethod({
```
morgsmccauley, line 111: At this point we'd actually be creating a new document rather that updating right? I know `updateRecoveryMethod` will create if it doesn't exist, curious if there is any reason to use this over `createRecoveryMethod`?

morgsmccauley: Also do we need to be concerned about `deleteRecoveryMethod` succeeding and this (creation) failing? should we `catch` errors thrown from this and add some specific logging to let CSA know something has gone wrong and will require action from an engineer? Maybe printing the original document would be useful here?

morgsmccauley (changes requested):
Looks good, can't wait to get this out :D 

Just a couple questions.

scripts/multisig.js
```diff
@@ -0,0 +1,140 @@
...
+}
+
+async function lookupRecoveryMethods(accountId) {
+    if (!isAccountValid(accountId)) {
+        console.error(`Invalid account ID ${accountId}`);
+        return;
+    }
+
+    const recoveryMethods = await listRecoveryMethodsByAccountId(accountId)
+        .map(({ detail, kind }) => ({ detail, kind }));
```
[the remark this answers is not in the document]
andy-haynes: I just like my helper methods 😄

scripts/multisig.js
andy-haynes: > At this point we'd actually be creating a new document rather that updating right?
Good point, I had the update in there before I realized that the library wouldn't delete the current document as part of updating the range key.

However since I made the change to support email-email transfers I think I will leave the update but make the delete contingent on whether the range key is updated.

> Also do we need to be concerned about `deleteRecoveryMethod` succeeding and this (creation) failing?
Not if it remains as an update I think. Good call on printing the original document upfront though, I'll add that.

scripts/multisig.js
```diff
@@ -0,0 +1,140 @@
...
+
+    if (!isEmail(desired)) {
+        console.error(`Invalid email: ${desired}. The 2FA method can only be set to email.`);
+        return;
+    }
+
+    const isCurrentPhone = !isEmail(current);
+    const currentMethodKind = isCurrentPhone ? TWO_FACTOR_AUTH_KINDS.PHONE : TWO_FACTOR_AUTH_KINDS.EMAIL;
+    const [sms2faMethod] = await listRecoveryMethodsByAccountId(accountId)
+        .filter(({ detail, kind }) => kind === currentMethodKind && (isCurrentPhone ? isEquivalentPhoneNumber(detail, current) : (detail === current)));
```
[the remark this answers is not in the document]
andy-haynes: Yeah this was a quick hack I added while testing the refactor and meant to clean up. It should now support transfers to either kind.

scripts/multisig.js
```diff
@@ -0,0 +1,140 @@
...
+function is2faRecoveryMethod(kind) {
+    return Object.values(TWO_FACTOR_AUTH_KINDS).includes(kind);
+}
+
+function isAccountValid(accountId) {
+    if (accountId.length === 64) {
+        return true;
+    }
+
+    if (process.env.NEAR_WALLET_ENV.startsWith('testnet')) {
```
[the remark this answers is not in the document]
andy-haynes: I think it probably is better as a CLI option but unfortunately anything referencing DB code needs it set at runtime. The instructions have the user specify `NEAR_WALLET_ENV` as a shell variable directly.

scripts/multisig.js
```diff
@@ -0,0 +1,164 @@
...
+    const isCurrentEmail = isEmail(current);
+    const isDesiredEmail = isEmail(desired);
+
+    if (!isDesiredEmail) {
+        console.error(`Invalid email: ${desired}. The 2FA method can only be set to email.`);
+        return;
+    }
+
+    const currentMethodKind = isCurrentEmail ? TWO_FACTOR_AUTH_KINDS.EMAIL : TWO_FACTOR_AUTH_KINDS.PHONE;
+    const [sms2faMethod] = await listRecoveryMethodsByAccountId(accountId)
```
morgsmccauley, line 87:
```suggestion
    const [current2faMethod] = await listRecoveryMethodsByAccountId(accountId)
```
This could also be `email` right?

morgsmccauley (approved):

PR #606, opened 2022-05-03
WAL-301 Create entry point for indexer service

morgsmccauley:
This PR creates a new entry point to serve **only** indexer related endpoints. This is an initial step for separating out the indexer queries to a standalone service and reducing the blast radius when these queries fail.

- feat: Duplicate `app.js` to create initial `indexer.js`
- feat: Reduce `indexer.js` to expose indexer endpoints only
- feat: Add package.json script to start indexer
- refactor: Use `NEAR_WALLET_ENV` to determine env instead of `WALLET_URL`
- refactor: Environment usage in `indexer.js`


package.json
```diff
@@ -5,6 +5,7 @@
   "main": "app.js",
   "scripts": {
     "start": "env $(sed 's/ # .*//' .env) supervisor app",
+    "start:indexer": "env $(sed 's/ # .*//' .env) supervisor indexer",
```
andy-haynes, line 8: Can we change this to `node indexer.js`? We don't need to watch the codebase or parse the `.env` file (which won't exist anyway). If the process does die then I think we want the container to exit.

For reference the `docker:start` command is what the ECS service currently uses as its entrypoint.

andy-haynes (changes requested):
Looks good, just a comment on the script

package.json
morgsmccauley: If we don't need to reference `supervisor` should we just do without a `package.json` script all together? The task definition could just include `node indexer.js`?

andy-haynes: Yup but it provides a convenient alias for the task definition to reference so that it doesn't need to be updated if we refactor something in this codebase. I feel like this repo should be providing the interface to start the indexer service and external consumers shouldn't need knowledge of the module structure to run it. So I'd prefer to have it even if it's functionally equivalent at the moment.

morgsmccauley: Good points, we probably still need `.env` parsing + `supervisor` for local dev so i've kept `start:indexer` the same and added `docker:start:indexer` which is just `node indexer.js`

andy-haynes (approved):

PR #607, opened 2022-05-04
chore: Move application code to `src/` directory

morgsmccauley:
Moves all application code under top-level `src/` directory. `test/` is unchanged and also lives in root dir which is why the file paths change within most of the test files.

- chore: move application code to `src/`
- chore: update file references to include `src/`
- chore: update package.json scripts to include `src/`


andy-haynes (approved):
LGTM

PR #616, opened 2022-05-23
feat: Bypass account ownership check on ledger/seed phrase recovery creation for non-existent implicit accounts

morgsmccauley:
This PR bypasses the account ownership check on creation of recovery methods for ledger (`/account/ledgerKeyAdded`) and seed phrases (`/account/seedPhraseAdded`) if the account:
1. Does not exist on-chain
1. Has an implicit account ID

For named accounts which exist on-chain there is no change, if the named account does not exist the middleware will return `403`.

This change is to support creation of zero-balance ledger/seed phrase accounts.

---

- refactor(near): Extract account existence check to own function
- feat(account): Skip ownership check if implicit account doesnt exist
- test(account): Test account ownership skip for ledger/seed phrase

### DIFF
--- a/middleware/near.js
+++ b/middleware/near.js
@@ -56,7 +56,7 @@ async function checkAccountOwnership(ctx, next) {
     return await next();
 }
 
-// TODO: near-api-js should have explicit accoutn existence check
+// TODO: near-api-js should have explicit account existence check
 async function getAccountExists(near, accountId) {
     try {
         await (await near.account(accountId)).state();
--- a/middleware/near.js
+++ b/middleware/near.js
@@ -56,6 +56,20 @@ async function checkAccountOwnership(ctx, next) {
     return await next();
 }
 
+// TODO: near-api-js should have explicit accoutn existence check
+async function getAccountExists(near, accountId) {
+    try {
+        await (await near.account(accountId)).state();
+    } catch (e) {
+        if (e.type === 'AccountDoesNotExist') {
+            return false;
+        }
+        throw e;
+    }
+
+    return true;
+}
+
 function createCheckAccountDoesNotExistMiddleware({ source, fieldName }) {
     if (source !== 'body' && source !== 'params') {
         throw new Error('invalid source for accountId provided');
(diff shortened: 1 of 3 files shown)


middleware/near.js
```diff
@@ -56,6 +56,37 @@ async function checkAccountOwnership(ctx, next) {
...
+        if (e.type === 'AccountDoesNotExist') {
+            return false;
+        }
+        throw e;
+    }
+
+    return true;
+}
+
+async function accountAuthMiddleware(ctx, next) {
```
morgsmccauley, line 73: Open to naming suggestions, this flow is hard to describe in a concise name 😅

middleware/near.js
```diff
@@ -56,6 +56,37 @@ async function checkAccountOwnership(ctx, next) {
...
+
+    if (!isImplicitAccount(accountId)) {
+        ctx.throw(403, `Named account ${accountId} does not exist`);
+    }
+
+    return next();
+}
+
+const isImplicitAccount = (accountId) =>
+    accountId && accountId.length === 64 && !accountId.includes('.');
```
morgsmccauley, line 88: Same check as we have on frontend

middleware/near.js
```diff
@@ -74,19 +105,11 @@ function createCheckAccountDoesNotExistMiddleware({ source, fieldName }) {
...
-        let remoteAccount = null;
-        try {
-            remoteAccount = await (await ctx.near.account(accountId)).state();
-        } catch (e) {
-            if (e.type === 'AccountDoesNotExist') {
-                return await next();
-            }
-            throw e;
-        }
-        if (remoteAccount) {
```
morgsmccauley, line 87: Removed this check as we should always have an `account` at this point, `near.account().state()` would throw otherwise - correct me if I am wrong though.

test/app.test.js
```diff
@@ -335,6 +337,30 @@ describe('app routes', function () {
             const account = await accountService.getAccount(accountId);
             expect(account).ok;
         });
+
+        it('returns 403 if named account does not exist', async () => {
+            const nonExistentNamedAccountId = testAccountHelper.buildTestAccountId();
```
morgsmccauley, line 342: This is essentially a UUID so there is a (low?) chance of it actually existing on-chain and this test therefore failing. This is the same risk we take when creating new accounts via `testAccountHelper.createNEARAccount()` though so is potentially fine?

Patrick1904 (approved):
Thank you @morgsmccauley! 🙏  LGTM.

andy-haynes (approved):

PR #637, opened 2022-07-21
feat: fromBlockTimestamp query param to support caching

esaminu:
This PR adds functionality to query a block timestamp range by adding a `fromBlockTimestamp` query param to the `/likelyTokens` and `/likelyNFTs` endpoint

### DIFF
--- a/indexer.js
+++ b/indexer.js
@@ -9,7 +9,9 @@ const {
     findStakingDeposits,
     findReceivers,
     findLikelyTokens,
+    findLikelyTokensFromBlock,
     findLikelyNFTs,
+    findLikelyNFTsFromBlock,
     findStakingPools,
     findAccountActivity,
 } = require('./src/middleware/indexer');
@@ -37,6 +39,8 @@ router.get('/account/:accountId/activity', findAccountActivity);
 router.get('/account/:accountId/callReceivers', findReceivers);
 router.get('/account/:accountId/likelyTokens', findLikelyTokens);
 router.get('/account/:accountId/likelyNFTs', findLikelyNFTs);
+router.get('/account/:accountId/likelyTokensFromBlock', findLikelyTokensFromBlock);
+router.get('/account/:accountId/likelyNFTsFromBlock', findLikelyNFTsFromBlock);
 router.get('/stakingPools', findStakingPools);
 
 app
--- a/src/middleware/indexer.js
+++ b/src/middleware/indexer.js
@@ -112,6 +112,10 @@ const findReceivers = async (ctx) => {
 
 const findLikelyTokens = async (ctx) => {
     const { accountId } = ctx.params;
+    const { fromBlockTimestamp = 0 } = ctx.query;
+
+    const latestBlockTimeStampQuery = 'select block_timestamp FROM blocks ORDER BY block_timestamp DESC LIMIT 1';
+    const { rows: [ { block_timestamp: latestBlockTimestamp } ] } = await pool.query(latestBlockTimeStampQuery);
 
     const received = `
         select distinct receipt_receiver_account_id as receiver_account_id
@@ -120,6 +124,8 @@ const findLikelyTokens = async (ctx) => {
             and action_kind = 'FUNCTION_CALL'
             and args->>'args_json' is not null
             and args->>'method_name' in ('ft_transfer', 'ft_transfer_call','ft_mint')
+            and receipt_included_in_block_timestamp <= $3
+            and receipt_included_in_block_timestamp > $4
     `;
 
     const mintedWithBridge = `
@@ -129,6 +135,8 @@ const findLikelyTokens = async (ctx) => {
             where action_kind = 'FUNCTION_CALL' and
                 receipt_predecessor_account_id = $2 and
                 args->>'method_name' = 'mint'
+                and receipt_included_in_block_timestamp <= $3
+                and receipt_included_in_block_timestamp > $4
         ) minted_with_bridge
         where account_id = $1
     `;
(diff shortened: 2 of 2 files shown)


frol (dismissed):
LGTM

src/middleware/indexer.js
```diff
@@ -136,21 +141,25 @@ const findLikelyTokens = async (ctx) => {
...
             and (args->>'method_name' like 'ft_%' or args->>'method_name' = 'storage_deposit')
+            and receipt_included_in_block_timestamp <= $3
+            and receipt_included_in_block_timestamp > $4
     `;
 
     const ownershipChangeEvents = `
         select distinct emitted_by_contract_account_id as receiver_account_id 
         from assets__fungible_token_events
         where token_new_owner_account_id = $1
+            and emitted_at_block_timestamp <= $3
```
morgsmccauley, line 152: `$3` is just the latest block timestamp, is there really any benefit in adding this?

src/middleware/indexer.js
```diff
@@ -136,21 +141,25 @@ const findLikelyTokens = async (ctx) => {
...
+    const { rows } = await pool.query([received, mintedWithBridge, calledByUser, ownershipChangeEvents].join(' union '), [accountId, BRIDGE_TOKEN_FACTORY_ACCOUNT_ID, lastBlockTimestamp, fromBlockTimestamp]);
+    return { rows, lastBlockTimestamp };
 };
 
-
-const findLikelyNFTs = async (ctx) => {
-    const { accountId } = ctx.params;
+const likelyNFTsFromBlock = async (fromBlockTimestamp = 0, accountId) => {
+    const latestBlockTimeStampQuery = 'select block_timestamp FROM blocks ORDER BY block_timestamp DESC LIMIT 1';
+    const { rows: [ { block_timestamp: lastBlockTimestamp } ] } = await pool.query(latestBlockTimeStampQuery);
```
morgsmccauley, line 162: Let's separate this out in to it's own function since it's being reused?

indexer.js
```diff
@@ -34,6 +36,8 @@ router.get('/account/:accountId/activity', findAccountActivity);
 router.get('/account/:accountId/callReceivers', findReceivers);
 router.get('/account/:accountId/likelyTokens', findLikelyTokens);
 router.get('/account/:accountId/likelyNFTs', findLikelyNFTs);
+router.get('/account/:accountId/likelyTokensFromBlock', findLikelyTokensFromBlock);
+router.get('/account/:accountId/likelyNFTsFromBlock', findLikelyNFTsFromBlock);
```
morgsmccauley, line 40:
```suggestion
router.get('/v2/account/:accountId/likelyTokens', findLikelyTokensFromBlock);
router.get('/v2/account/:accountId/likelyNFTs', findLikelyNFTsFromBlock);
```
Maybe we should expose these as `v2` endpoints instead, the logic between them is essentially the same so would be good to eventually deprecate the old ones. Thoughts?

morgsmccauley (changes requested):

src/middleware/indexer.js
esaminu: hmm we want to return the last block in the response so the client can query from that next time so I guess it's in case a block gets added during the queries? @frol can confirm but it's important we don't miss any blocks here since FTs / NFTs can be minted in those blocks.

indexer.js
esaminu: Interesting idea but it's a bit redundant with the `version` param we're returning on the endpoint. I lean on the side of `version` param since I don't want clients to have to update their requests every time the query to the indexer changes where they can instead just rescan with `fromBlockTimestamp` 0 if they detect a version that's different from their cache. I agree that we should deprecate the old ones though. Is there a better way to support this rescan functionality?

morgsmccauley: Ah I see the reason for `version` now, thought it was more related to the endpoint, rather than the internal behavior. Let's leave as is 👍🏽

src/middleware/indexer.js
```diff
@@ -162,18 +174,67 @@ const findLikelyNFTs = async (ctx) => {
...
-    const { rows } = await pool.query([ownershipChangeFunctionCalls, ownershipChangeEvents].join(' union '), [accountId]);
+    const { rows } = await pool.query([ownershipChangeFunctionCalls, ownershipChangeEvents].join(' union '), [accountId, lastBlockTimestamp, fromBlockTimestamp]);
+    return { rows, lastBlockTimestamp };
+};
+
+const findLikelyTokens = async (ctx) => {
+    const { accountId } = ctx.params;
+
+    const { rows } = await likelyTokensFromBlock(0, accountId);
     ctx.body = rows.map(({ receiver_account_id }) => receiver_account_id);
```
andy-haynes, line 197: Are there any callers of `likelyTokensFromBlock` or `likelyNFTsFromBlock` that _don't_ need to map this data to return a list of strings? It seems like this `.map` should be happening in the `likely*FromBlock` methods if nothing is relying on `rows` being a list of objects. Individual callers shouldn't be responsible for doing the `.map` if that's the expected shape of the returned value.

src/middleware/indexer.js
```diff
@@ -162,18 +174,67 @@ const findLikelyNFTs = async (ctx) => {
             and action_kind = 'FUNCTION_CALL'
             and args->>'args_json' is not null
             and args->>'method_name' like 'nft_%'
+            and receipt_included_in_block_timestamp <= $2
```
andy-haynes, line 177: `actions_receipt_actions` has an index on `args->'args_json'->>'receiver_id'` but it doesn't reference `receipt_included_in_block_timestamp`. Can we see about getting the timestamp added to that index so that it's conducive to range queries? At the very least we should compare query plans between the current calls and this range query to rule it out as an issue - we don't want our well-intentioned change causing more DB contention.

src/middleware/indexer.js
```diff
@@ -162,18 +174,67 @@ const findLikelyNFTs = async (ctx) => {
...
+            and receipt_included_in_block_timestamp > $3
     `;
 
     const ownershipChangeEvents = `
         select distinct emitted_by_contract_account_id as receiver_account_id 
         from assets__non_fungible_token_events
         where token_new_owner_account_id = $1
+            and emitted_at_block_timestamp <= $2
```
andy-haynes, line 185: Same comment here re: indexes. If we can append the timestamp to the existing index that's great, otherwise let's at least get some assurance that the change in performance is negligible.

andy-haynes (changes requested):
Looking good, minor comments inline but the bigger concern here is that our new timestamp range checks don't degrade query performance. The queries affected by this PR seem to more or less line up with the indexes on the referenced tables as they are currently used, but very few of those indexes reference the timestamps which are now parameterized.

Would you please mind generating the before and after query plans for both of these queries and adding them as a comment on this PR so that we have a record of the changes that's associated with the code change? If the changes are negligible then I think it's OK to ship as is, but if we see a dramatic difference in the query plans generated then we should explore updating the indexes currently utilized. Hopefully the changes are negligible but unfortunately we don't have a good track record with stable query plans 🙂

andy-haynes:
I went ahead and compared query plans - images are before (left) and after (right), always comparing the current query with the new one. Overall LGTM, changes in plan structure seem small overall and the differences in metrics appear pretty marginal. Hopefully no problems with plans changing 🤞 

`likelyNFTsFromBlock`

Small increase in planning/execution in the before/after for the initial case:
image

Bigger increase in planning/execution in the before/after for the diff case:
image

`likelyTokensFromBlock`

Very small increase in plan time/execution cost but decreased execution time in the before/after for the initial case:
image

Large apparent decrease in overall cost/execution time in the before/after for the diff case:
image

src/middleware/indexer.js
frol: As @esaminu explained, this clause prevents potential race condition.

src/middleware/indexer.js
```diff
@@ -110,8 +115,8 @@ const findReceivers = async (ctx) => {
     ctx.body = rows.map(({ receiver_account_id }) => receiver_account_id);
 };
 
-const findLikelyTokens = async (ctx) => {
-    const { accountId } = ctx.params;
+const likelyTokensFromBlock = async (fromBlockTimestamp = 0, accountId) => {
```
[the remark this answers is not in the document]
esaminu: No strong opinions here at all. Changed to object and removed default param

src/middleware/indexer.js
esaminu: I'd prefer to leave this as is since this can be generalized to other uses aside from listing `receiver_account_id`s. I did the same thing for `findLastBlockByTimestamp`. It's good to have generic methods that query pg and return rows here imo.

esaminu:
Thanks for raising these concerns and then investigating @andy-haynes! ❤️

src/middleware/indexer.js
andy-haynes: Gotcha, good point 👍

andy-haynes (approved):
LGTM, nice work!

morgsmccauley (approved):

PR #638, opened 2022-07-25
ci: Migrate ECS deployments to Github actions

morgsmccauley:
This PR migrates contract-helper & indexer deployments from Jenkins to Github Actions. The main drivers behind this are:
1. Actions are serverless meaning no operational burden
2. Setting up the deployment pipeline on forks will be much easier

Currently, the Jenkins pipeline does not wait for ECS service stability before progressing to the next step, it essentially fires and forgets, meaning individual service deployments could fail without the pipeline knowing. I've added these checks in to this pipeline so each step will wait for the ECS deployment to finish. Everything else should have parity with the existing.


MaximusHaximus (approved):
:+1:

PR #642, opened 2022-09-13
Fix: set validators cache time to 1 hour

PinkiNice:
According to the node-cache documentation: https://www.npmjs.com/package/node-cache

Current TTL time is set incorrectly (60 seconds instead of wanted 1 hour); This PR fixes this issues


roshaans (approved):
LGTM

morgsmccauley (approved):

PR #643, opened 2022-09-13
feat: Add error listener to postgres pool

morgsmccauley:
This PR adds an error listener to `pg.Pool` to avoid the error bubbling up and causing the node process to crash. The error is caught and logged. `Pool` should handle re-connection under the hood.


MaximusHaximus (approved):
:+1: :shipit:

PR #645, opened 2022-09-21
feat(db): local DynamoDB for development

andy-haynes:
This PR extends the current usage of the local DynamoDB used for testing to be used during development. When `NODE_ENV` is set to `development` the `dynamo` module is overridden to use a local DynamoDB instance which has been started by the `yarn start` command.

The contents of the local DynamoDB instance are persisted in a `shared-local-instance.db` file. To clear the database users may delete this file and restart the service.


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "Micro-service used by NEARStudio to deploy smart contracts.",
   "main": "app.js",
   "scripts": {
-    "start": "env $(sed 's/ # .*//' .env) supervisor app | bunyan",
+    "start": "node scripts/run-development-dynamo.js & env $(sed 's/ # .*//' .env) supervisor app | bunyan",
```
andy-haynes, line 7: Since we don't use `yarn start` in production environments this should be fine here.

local_dynamo.js
```diff
@@ -0,0 +1,49 @@
...
+        IdentityVerificationMethod,
+        RecoveryMethod,
+    ].map((model) => model.createTableAsync()))
+        .catch(() => {});
+
+    return {
+        port,
+        terminateLocalDynamo: () => {
+            console.warn(`Stopping local DynamoDB instance on port ${port}`);
+            dynamoLocal.stop(port);
```
morgsmccauley, line 39:
```suggestion
            await dynamoLocal.stop(port);
```
Feels like the `console.warn` below would be incorrect if we didn't `await` here.

morgsmccauley (approved):
Looks good, just one question.

local_dynamo.js
andy-haynes: Turns out it was a synchronous method the whole time: https://github.com/rynop/dynamodb-local/blob/master/index.js#L100

I'll remove the `Stopped local...` logging below to make it a little less verbose though

morgsmccauley (approved):

PR #653, opened 2022-10-19
feat: configurable dynamodb-local instance

andy-haynes:
This PR makes it possible to override the hardcoded host/ports for the local/test DynamoDB instance using environment variables. The defaults will continue to reference the original hardcoded values.

### DIFF
--- a/local_dynamo.js
+++ b/local_dynamo.js
@@ -8,13 +8,14 @@ const EmailDomainBlacklist = require('./src/db/schemas/email_domain_blacklist');
 const IdentityVerificationMethod = require('./src/db/schemas/identity_verification_method');
 const RecoveryMethod = require('./src/db/schemas/recovery_method');
 
-const LOCAL_DYNAMODB_PORT = 7877;
-const TEST_DYNAMODB_PORT = 7879;
+const LOCAL_DYNAMODB_HOST = process.env.LOCAL_DYNAMODB_HOST || 'localhost';
+const LOCAL_DYNAMODB_PORT = process.env.LOCAL_DYNAMODB_PORT || 7877;
+const TEST_DYNAMODB_PORT = process.env.TEST_DYNAMODB_PORT || 7879;
 
 function overrideLocalDynamo({ port } = { port: LOCAL_DYNAMODB_PORT }) {
     dynamo.documentClient(new DocumentClient({
         convertEmptyValues: true,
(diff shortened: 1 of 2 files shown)


gbouv (dismissed):
Tested it today and I am able to run DynamoDB on a separate Docker container. LGTM

local_dynamo.js
```diff
@@ -8,13 +8,20 @@ const EmailDomainBlacklist = require('./src/db/schemas/email_domain_blacklist');
...
 const RecoveryMethod = require('./src/db/schemas/recovery_method');
 
-const LOCAL_DYNAMODB_PORT = 7877;
-const TEST_DYNAMODB_PORT = 7879;
+const LOCAL_DYNAMODB_HOST = process.env.LOCAL_DYNAMODB_HOST || 'localhost';
+const LOCAL_DYNAMODB_PORT = process.env.LOCAL_DYNAMODB_PORT || 7877;
+const TEST_DYNAMODB_PORT = process.env.TEST_DYNAMODB_PORT || 7879;
+
+function createTables() {
+    const models = [Account, EmailDomainBlacklist, IdentityVerificationMethod, RecoveryMethod];
```
gbouv, line 16: Can we at least add a check here such that if the tables already exist, skip the creation?

andy-haynes: > Can we at least add a check here such that if the tables already exist, skip the creation?

It fails silently if the tables already exist, it's just a simpler implementation where more complexity is not required. This entire feature is essentially written to 1) run integration tests and 2) mitigate errors in the local dev environment due to no DB running (NB these errors do not affect contract helper usage in a practical way). I've yet to encounter any contributors working on our data layer, which is slated for deprecation in any case.

> And then, in an ideal world, call this from overrideLocalDynamo as well such that the contract helper is smart enough and do not rely on tribal knowledge that tables need to be seeded before it starts

It's not the responsibility of the application to prepare its own database. The expectation is that the infrastructure required by the target environment has been correctly provisioned/initialized. That's why it's handled in the local instances started to satisfy `yarn start` and `yarn test`.

It's also worth pointing out that this initialization is pretty trivial compared to the previous Postgres implementation, which required setting up database users and running ORM migrations (which automatically created missing tables). This would have been performed on a running Postgres instance and not within the application itself.

gbouv (approved):

PR #655, opened 2022-11-03
fix: external local dynamo init

andy-haynes:
This PR configures the local DynamoDB environment prior to table initialization.


gbouv (approved):
Tested it locally and it seems it fixes the issue. Thanks a lot @andy-haynes 👏

thisisjoshford:
Thank you so much for working with @gbouv and @mieubrisse to get local development w/ Kurtosis unblocked! 🙏🏼

PR #662, opened 2023-02-14
fix: remove reference to assets__fungible_token_events

andy-haynes:
This PR removes a reference from the indexer queries to a soon-to-be-deprecated table (`assets__fungible_token_events`). As this table will be removed entirely on March 1st 2023, it must be removed from Contract Helper to prevent queries from breaking.

This table is only referenced from within a subquery used to find fungible tokens. Removing this query will reduce the efficacy/reliability of fungible tokens query, in particular those FTs for which the only record of ownership is a transfer of ownership from another account.

https://github.com/near/near-indexer-for-explorer/discussions/351

### DIFF
--- a/src/middleware/indexer.js
+++ b/src/middleware/indexer.js
@@ -156,15 +156,7 @@ const likelyTokensFromBlock = async ({ fromBlockTimestamp, accountId }) => {
             and receipt_included_in_block_timestamp > $4
     `;
 
-    const ownershipChangeEvents = `
-        select distinct emitted_by_contract_account_id as receiver_account_id 
-        from assets__fungible_token_events
-        where token_new_owner_account_id = $1
-            and emitted_at_block_timestamp <= $3
-            and emitted_at_block_timestamp > $4
-    `;
-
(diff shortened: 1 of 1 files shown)


telezhnaya (approved):
Thank you for your quick reaction!
I'm OK with this as the temporary solution, but, for the next PR, I highly recommend dropping all the FT-related logic and just using [Enhanced API](https://console.pagoda.co/apis?tab=enhancedApi#/), everything is already collected for you

andy-haynes:
Thank you for the review @telezhnaya!

Contract Helper is more or less in maintenance mode at the moment so it is unlikely that transitioning to Enhanced API will be an immediate priority for this service. When we decommission `wallet.near.org` then the indexers here should be deprecated to drive adoption of Enhanced API by those who continue to rely on NEAR Contract Helper for wallet-related indexer data.

However if the long-term plan ends up keeping Contract Helper as the de-facto Free Wallet Indexer Service then it would make sense for us to replace it here rather than deprecate.

Issue #684, opened 2024-04-04
findAccountsByPublicKey: Change the datasource from ExplorerDB to QueryAPI


eduohe:
Indexer for explorer Postgres DB will be deprecated on Apr 30th, the endpoint  [findAccountsByPublicKey](https://github.com/near/near-contract-helper/blob/master/src/middleware/indexer.js#L119)   needs to use the new datasource offered by QueryAPI.

This issue is related to https://github.com/near/queryapi/issues/322

pkudinov:
@eduohe when are we planning to release this to prod? I will change this ticket to "In Review". Please mark as done when this is in production this week

PR #685, opened 2024-04-17
Migrate findAccountsByPublicKey endpoint to QueryAPI

eduohe:
Closes https://github.com/near/near-contract-helper/issues/684

Related to https://github.com/near/queryapi/issues/325

The response continues the same using QueryAPI

```
curl -XGET http://localhost:3050/publicKey/ed25519:FXFhPD2aYw6YZB9pa9UYJNkJgKGuWDsJvDH13B5NjehL/accounts
```
QueryAPI response:
Screenshot 2024-04-17 at 2 39 43 PM

Explorer DB record:
Screenshot 2024-04-17 at 2 41 47 PM

### DIFF
--- a/src/middleware/indexer.js
+++ b/src/middleware/indexer.js
@@ -121,30 +121,48 @@ const findAccountActivity = async (ctx) => {
 const findAccountsByPublicKey = async (ctx) => {
     try {
         const { publicKey } = ctx.params;
+        let accounts = [];
 
-        const response = await fetch(GRAPHQL_URL, {
-            method: 'POST',
-            headers: {
-                'content-type': 'application/json',
-                'x-hasura-role': 'dataplatform_near'
-            },
-            body: JSON.stringify({
-                query: `
-                    query access_keys_v1_by_public_key {
-                        dataplatform_near_access_keys_v1_access_keys_v1(
-                        where: {public_key: {_eq: \"${publicKey}\"}}
-                        ) {
-                        account_id
+        if (NEAR_WALLET_ENV.startsWith('mainnet')) {
+            const response = await fetch(GRAPHQL_URL, {
+                method: 'POST',
+                headers: {
+                    'content-type': 'application/json',
+                    'x-hasura-role': 'dataplatform_near'
+                },
+                body: JSON.stringify({
+                    query: `
+                        query access_keys_v1_by_public_key {
+                            dataplatform_near_access_keys_v1_access_keys_v1(
+                            where: {public_key: {_eq: \"${publicKey}\"}}
+                            ) {
+                            account_id
+                            }
                         }
-                    }
-                `,
-                operationName: 'access_keys_v1_by_public_key'
-            }),
-        });
+                    `,
+                    operationName: 'access_keys_v1_by_public_key'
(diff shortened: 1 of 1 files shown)


src/middleware/indexer.js
```diff
@@ -1,9 +1,11 @@
 const { Pool } = require('pg');
 const Cache = require('node-cache');
 const bunyan = require('bunyan');
+const fetch = require('node-fetch');
 
 const {
     BRIDGE_TOKEN_FACTORY_ACCOUNT_ID = 'factory.bridge.near',
+    GRAPHQL_URL = 'https://near-queryapi.api.pagoda.co/v1/graphql',
```
andy-haynes, line 8: Does this handle `testnet` queries as well?

src/middleware/indexer.js
```diff
@@ -119,15 +121,30 @@ const findAccountActivity = async (ctx) => {
...
+            method: 'POST',
+            headers: {
+                'content-type': 'application/json',
+                'x-hasura-role': 'dataplatform_near'
+            },
+            body: JSON.stringify({
+                query: `
+                    query access_keys_v1_by_public_key {
+                        dataplatform_near_access_keys_v1_access_keys_v1(
+                        where: {public_key: {_eq: \"${publicKey}\"}}
```
andy-haynes, line 135: Doesn't look like the `"` need to be escaped here

src/middleware/indexer.js
morgsmccauley: Ah that's a good point, it does not. We may need to parameterise this in ECS, and use ExplorerDB in testnet, QueryAPI doesn't support testnet yet.

eduohe: thanks for the heads up, I changed to have a fallback for non-mainnet environments or if no results from graphql

src/middleware/indexer.js
eduohe: thanks, removed

src/middleware/indexer.js
andy-haynes: Do you actually want a fallback for mainnet queries that return no results from graphql? If the DBs are actually going to be unavailable after April 30th then I imagine you'd want to know sooner than later that there's a disparity in what QueryAPI is returning. Also saves you the effort of needing to come back in < 2 weeks to remove the fallback 🙂 

Are the testnet indexer DBs also going to be unavailable after April 30th? It's sounding like there will be an indeterminate gap in coverage for testnet. I don't know how severe the real impact of this would be, it's probably safe to assume there isn't a lot of account recovery, staking, etc. coming from testnet environments. It would almost certainly break parity for wallets supporting testnet but it might not be a dealbreaker in the short term.

eduohe: Good questions, yes we want to keep the fallback, April 30th is the desired deadline, but there are other endpoints that needs to be migrated before we turn off the DB instance.

Testnet indexer DBs will stay, there is no deadline or workstream yet to deprecate them.

morgsmccauley (approved):
LGTM from my side :)

andy-haynes (approved):

